PR #24, opened 2023-09-09
feat: test gotrue when app start, use state for gotrue client, use go…

speed2exe:
- auto create user after login, which will trigger SQL that generates workspaces
- when cloud-server starts, it requires gotrue to be available as well (just like redis and postgres)
- server's gotrue client to Data<State>, along with redis and postgres


migrations/20230312043023_user.sql
```diff
@@ -3,9 +3,9 @@
 CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
 -- user table
 CREATE TABLE IF NOT EXISTS af_user (
-    uid BIGSERIAL PRIMARY KEY,
-    email TEXT NOT NULL DEFAULT '' UNIQUE,
-    password TEXT NOT NULL,
+    uid UUID PRIMARY KEY, -- related to gotrue user when authenticated
```
appflowy, line 6: The uid must be big int. I used it in appflowy applicaiton. How about create a new column: uuid

build/run_local_server.sh
```diff
@@ -32,4 +32,7 @@ fi
 # Kill any existing instances
 pkill -f appflowy_cloud || true
 
+# Run the migrations
+cargo sqlx database create && cargo sqlx migrate run && cargo sqlx prepare
```
appflowy, line 36: when performing docker compose. The migration should auto run by this line of code:

`- ./migrations:/docker-entrypoint-initdb.d`

So, is it necessary to create database and run the migrations?

libs/storage/src/entities.rs
```diff
@@ -19,3 +23,38 @@ impl CreateCollabParams {
...
+  pub database_storage_id: Option<sqlx::types::uuid::Uuid>,
+  pub owner_uid: Option<i64>,
+  pub created_at: Option<DateTime<Utc>>,
+  pub workspace_type: i32,
+  pub deleted_at: Option<DateTime<Utc>>,
+  pub workspace_name: Option<String>,
+}
+
+#[derive(Debug, sqlx::FromRow)]
+pub struct AfUserProfileView {
```
appflowy, line 39: Would it be better if replacing the AF instead of Af? Maybe we can document it and stick to that style.

speed2exe: yeah we can do that, will make the changes

build/run_local_server.sh
speed2exe: let's say you change some SQL file, the schema of DB will change, but it will not update the .sqlx files

migrations/20230312043023_user.sql
speed2exe: I've made the changes to use the existing BIGINT for user id

build/run_local_server.sh
appflowy: Got it.

So, it may just need to run : `cargo sqlx prepare`

PR #35, opened 2023-09-13
feat: support external oauth

speed2exe:


libs/shared-entity/src/error.rs
```diff
@@ -101,9 +106,27 @@ impl From<serde_json::Error> for AppError {
   }
 }
 
+impl From<opener::OpenError> for AppError {
```
appflowy, line 109: You can use #[cfg(feature = "cloud")] to isolate the dependencies from client_api crate

speed2exe: I see.  I have made the changes

PR #57, opened 2023-09-18
Nginx

speed2exe:
Add nginx component to compose
- exposes ports with both SSL and non-SSL from host


nginx/ssl/certificate.crt
speed2exe, line 1: certificates are not real ones

appflowy: Got it

PR #87, opened 2023-10-03
Collab redis cache

speed2exe:


libs/database/src/collab.rs
```diff
@@ -93,266 +91,118 @@ impl Default for StorageConfig {
...
+      pg_pool,
+      redis_client,
+      config,
+    }
   }
 }
 
 #[async_trait]
-impl CollabStorage for CollabPostgresDBStorageImpl {
+impl CollabStorage for CollabDatabaseStorageImpl {
```
appflowy, line 116: Create a new implementation of CollabStorage and keep the CollabPostgresDBStorageImpl.

libs/database/src/collaborate.rs
```diff
@@ -0,0 +1,306 @@
+use anyhow::Context;
```
appflowy, line 1: collab.rs is used to collab object. Maybe we can keep that name.

PR #119, opened 2023-10-16
feat: added oauth login for admin

speed2exe:


docker-compose.yml
```diff
@@ -60,7 +60,8 @@ services:
     depends_on:
       - postgres
     environment:
-      - GOTRUE_SITE_URL=appflowy-flutter://                           # redirected to AppFlowy application
```
appflowy, line 63: I think the appflowy-flutter can't be changed. It is also set in the frontend application.

PR #135, opened 2023-10-23
Feat/user invite

speed2exe:
- non admin user can invite other user
- an email will be sent to the invited user, containing a login link
- added login for github and discord
- added ability for users to navigate to app from web
- added pgadmin, minio and portainer quick links in the admin page


admin_frontend/src/web_api.rs
```diff
@@ -21,23 +25,77 @@ pub fn router() -> Router<AppState> {
...
+) -> Result<WebApiResponse<WebApiOpenAppResponse>, WebApiError<'static>> {
+  let access_token_resp = state
+    .gotrue_client
+    .token(&Grant::RefreshToken(RefreshTokenGrant {
+      refresh_token: session.refresh_token.to_owned(),
+    }))
+    .await?;
+
+  let app_sign_in_url = format!(
+      "appflowy-flutter://login-callback#access_token={}&expires_at={}&expires_in={}&refresh_token={}&token_type={}",
```
appflowy, line 60: The `appflowy-flutter://login-callback` requires documentation to explain its usage. One issue to note is that the URL will become invalid if we alter the schema. For now, we can proceed with it.

admin_frontend/templates/components/invite.html
```diff
@@ -0,0 +1,43 @@
...
+    <tr>
+      <td></td>
+      <td style="text-align: right">
+        <button class="button cyan" id="inviteBtn">Invite</button>
+      </td>
+    </tr>
+  </table>
+
+  <script>
+    document.getElementById("inviteBtn").addEventListener("click", function () {
```
appflowy, line 23: If we choose not to use any web frameworks (like React) here, we may need to refactor the HTML

admin_frontend/src/web_api.rs
speed2exe: I'll add some comments to clarify this part

admin_frontend/templates/components/invite.html
speed2exe: Most of the HTML are components and can be reused when necessary to keep the code dry. Is there a specific part you are referring to?

appflowy: No, it's just a concern

PR #141, opened 2023-10-27
Metrics

speed2exe:
- [x] Initial Prometheus Client boilercode
- [x] Added `/metrics` endpoint for prometheus to scrap data
- [x] add metrics for request count for each path
- [x] configure prometheus to scrap data from server
- [x] add histogram metrics for request duration for each path
- [x] add custom appflowy dashboard to grafana and save as json config file


src/application.rs
```diff
@@ -106,6 +113,7 @@ pub async fn run(
           .cookie_name(HEADER_TOKEN.to_string())
           .build(),
       )
+      .wrap(MetricsMiddleware)
```
appflowy, line 116: Maybe move to line 110 (under the IdentityMiddleware). IIRC, the middleware are invoked as reverser order.

PR #149, opened 2023-11-06
fix: restartable server when autoconfirm is true

speed2exe:
- This change will better support self-hosting in the condition that self hoster does not set up SMTP or any form of OAuth.
- Previously, restarting a server while having `GOTRUE_MAILER_AUTOCONFIRM=true`, will cause an error. With these changes, the server will check if admin user is already registered, skipping admin user creation ifalready created.


src/application.rs
```diff
@@ -213,26 +213,55 @@ async fn setup_admin_account(
...
+          gotrue_entity::dto::SignUpResponse::NotAuthenticated(user) => user,
+        }
+      };
+      match admin_user.role.as_str() {
+        "supabase_admin" => {
+          tracing::info!("Admin user already created and set role to supabase_admin");
+          Ok(())
+        },
+        _ => {
+          let user_id = admin_user.id.parse::<uuid::Uuid>().unwrap();
```
appflowy, line 244: Safety to unwrap?

speed2exe: fixed

PR #166, opened 2023-11-14
feat: Add cloudflare tunnel

ThanatosDi:
Use cloudflare tunnel to protect Appflowy Cloud so that the server does not need to expose the port to avoid port scanning

1. CloudFlare zero trust setting create a tunnel
  image

2. Get tunnel token
  image

3. put token to env
  ```bash
  CLOUDFLARE_TUNNEL_TOKEN=eyJhIjoiOWRiODVlM2Nm...
  ```
4. start services
  ```bash
  docker compose up -d 
  ```

5. Setting cloudflare tunnel public host
  image

### DIFF
--- a/dev.env
+++ b/dev.env
@@ -66,3 +66,6 @@ PORTAINER_PASSWORD=password1234
 # Grafana Dashboard
 GF_SECURITY_ADMIN_USER=admin
 GF_SECURITY_ADMIN_PASSWORD=password
+
+# Cloudflare tunnel token
+CLOUDFLARE_TUNNEL_TOKEN=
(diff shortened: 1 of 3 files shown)


appflowy:
Thank you for your contribution! The `docker-compose-dev` is only used during development. Therefore, I think we don't need the `tunnel` in `docker-compose-dev`.

ThanatosDi:
> Thank you for your contribution! The `docker-compose-dev` is only used during development. Therefore, I think we don't need the `tunnel` in `docker-compose-dev`.

OK, then I remove tunnel from `docker-compose-dev`

Jiraffe7:
The changes look good to me. I was able to run the tests, successfully accessing the server through cloudflare tunnel.

PR #206, opened 2023-12-12
fix: update the S3 env vars in docker-compose.yml to not have `AWS` in them

jessebot:
Thanks to https://github.com/AppFlowy-IO/AppFlowy-Cloud/issues/204#issuecomment-1851285350 I was able to test using these env vars and Appflowy-Cloud is able to initialize it's app state using the minio bucket I specified.

This PR updates those environment variables in the docker-compose.yml file to help others.


speed2exe:
@jessebot thanks for the PR, but CI is failing, I'll look into it and fix it before the merge.

PR #224, opened 2023-12-18
feat: use only env var for appflowy cloud

speed2exe:
- Use pure env config instead of configuration files
- update appflowy_cloud rust build toolchain


src/config/config.rs
```diff
@@ -101,55 +70,73 @@ impl DatabaseSetting {
...
+// Default values favor local development.
+pub fn get_configuration() -> Result<Config, anyhow::Error> {
+  let config = Config {
+    app_env: get_env_var("APPFLOWY_ENVIRONMENT", "local").parse()?,
+    db_settings: DatabaseSetting {
+      pg_conn_opts: PgConnectOptions::from_str(&get_env_var(
+        "APPFLOWY_DATABASE_URL",
+        "postgres://postgres:password@localhost:5433/postgres",
+      ))?,
+      require_ssl: get_env_var("APPFLOWY_DATABASE_REQUIRE_SSL", "false").parse()?,
```
appflowy, line 91: how about making these string as const &str

speed2exe: you mean the ones wrapped in ""? isn't already static lifetime &str? or are you referring the the result from `get_env_var`?

appflowy: Never mind. It's ok

PR #249, opened 2024-01-08
chore: compress collab data

appflowy:
Compress collab data when posting from AppFlowy Application

### DIFF
--- a/libs/client-api/src/http.rs
+++ b/libs/client-api/src/http.rs
@@ -413,14 +413,20 @@ impl Client {
     device_id: &str,
     msg: Message,
   ) -> Result<(), AppResponseError> {
+    let payload = msg.into_data();
+    let compressed_payload = compress(
+      &payload,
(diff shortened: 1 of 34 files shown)


libs/client-api/src/http.rs
```diff
@@ -52,7 +55,35 @@ use gotrue_entity::dto::SignUpResponse::{Authenticated, NotAuthenticated};
...
+    self.compression_buffer_size = compression_buffer_size;
+    self
+  }
+}
+
+impl Default for ClientConfiguration {
+  fn default() -> Self {
+    Self {
+      compression_quality: 10,
+      compression_buffer_size: 10240,
```
speed2exe, line 82: if these variables are changed after deployment and server restarts, will the data become invalid?

appflowy: Nope. Each request will carry the header values that contains the buffer size. The buffer size will be used to decompress the data.

PR #263, opened 2024-01-18
Update minio root user & pass variable to match DEPLOYMENT.MD & deploy.env

dmasiero:
This change allows for following variables found in lines 94 and 95 of DEPLOYMENT.MD .env example and lines 58 and 59 of deploy.env - Without this change, deployment easily deploys with default user/pass of minioadmin/minioadmin


speed2exe:
@dmasiero thank you for noticing and the PR!

PR #278, opened 2024-01-29
feat: client middleware for auto refresh

speed2exe:


libs/client-api/src/native/retry.rs
```diff
@@ -1,68 +1,71 @@
...
+// impl RefreshTokenAction {
+//   pub fn new(token: Arc<RwLock<ClientToken>>, gotrue_client: gotrue::api::Client) -> Self {
+//     Self {
+//       token,
+//       gotrue_client: Arc::new(gotrue_client),
+//     }
+//   }
+// }
+//
+// impl Action for RefreshTokenAction {
```
appflowy, line 25: How to auto retry if refresh fail?

libs/gotrue/src/api.rs
```diff
@@ -11,15 +11,16 @@ use gotrue_entity::dto::{
 use gotrue_entity::error::{GoTrueError, GoTrueErrorSerde, GotrueClientError};
 use gotrue_entity::sso::{SSOProvider, SSOProviders};
 use infra::reqwest::{check_response, from_body, from_response};
+use reqwest_middleware::ClientWithMiddleware;
 
 #[derive(Clone)]
 pub struct Client {
-  client: reqwest::Client,
+  client: ClientWithMiddleware,
```
appflowy, line 18: How about we design a custom struct that encapsulates the ClientWithMiddleware? This approach would simplify switching to a different client in the future. Maybe it's easier to do more optimization without changing the interface.

libs/client-api/src/native/http_native.rs
```diff
@@ -122,24 +121,43 @@ impl Client {
...
+    let mut current_retry_count = 0;
+    let max_retry_count = 4;
 
-        // If the error is an OAuth error, unset the token.
-        if err.is_unauthorized() {
-          self.token.write().unset();
-        }
-        Err(err.into())
-      },
+    let refresh_token = self
```
appflowy, line 127: I favor using the retry action because it is self-contained and allows for straightforward configuration adjustments, such as changing the strategy.

libs/client-api/src/native/http_native.rs
```diff
@@ -122,24 +121,43 @@ impl Client {
...
+
+      match self.gotrue_client.token(&grant_req).await {
+        Ok(new_token) => {
+          self.token.write().set(new_token);
+          return Ok(());
+        },
+        Err(err) => {
+          warn!("refresh token failed: {}", err);
+          if err.is_network_error() {
+            std::thread::sleep(Duration::from_secs(2));
```
appflowy, line 150: You might want to use tokio::time::sleep instead of std::thread::sleep. They are different.

> std::thread::sleep
Synchronously blocks the current thread for the specified duration (in this case, 2 seconds).
During the sleep period, the thread is completely inactive and does no work.

> tokio::time::sleep
While the task is sleeping, the Tokio runtime can run other tasks on the same thread.

libs/client-api/src/native/retry.rs
speed2exe: does current implementation and keeps waiting forever? there's issues integrating with ClientWithMiddleware. Not sure if there's another way.

libs/gotrue/src/api.rs
speed2exe: make sense

libs/client-api/src/native/http_native.rs
speed2exe: got it

libs/client-api/src/native/http_native.rs
speed2exe: I'll try to see if I can integrate it back into it

libs/client-api/src/native/retry.rs
appflowy: Nope. That depends on the strategy you use.

image

appflowy:
@speed2exe Can we close this PR?

PR #345, opened 2024-02-23
Refactor snapshot

appflowy:


src/biz/collab/metrics.rs
```diff
@@ -0,0 +1,33 @@
+use prometheus_client::metrics::gauge::Gauge;
+use prometheus_client::registry::Registry;
+use std::sync::atomic::AtomicU64;
+
+
+#[derive(Clone)]
+pub struct CollabMetrics {
+  success_write_snapshot_ratio: Gauge<f64, AtomicU64>,
```
speed2exe, line 8: for this i think it's better to have 2 counters, `total success write` and `total fail write`. then in the dashboard you can use `rate` function in the dashboard to let prometheus do the calculation based on the time.

appflowy: Got it!

PR #362, opened 2024-03-04
Workspace invite

speed2exe:
- Deprecate direct adding members to workspace.
- In order to add members to workspace, owners will sent an invitation. After receiver accepts invitation, he/she will then be added to the workspace


src/biz/workspace/ops.rs
```diff
@@ -131,6 +137,60 @@ pub async fn open_workspace(
...
+      &mut txn,
+      workspace_id,
+      inviter,
+      invitation.email.as_str(),
+      invitation.role,
+    )
+    .await?;
+  }
+
+  txn.commit().await?;
```
appflowy, line 179: how about providing the context("") here?

libs/database-entity/src/dto.rs
```diff
@@ -481,4 +481,35 @@ pub struct AFWorkspaceMember {
...
+  pub workspace_id: Uuid,
+  pub workspace_name: Option<String>,
+  pub inviter_email: Option<String>,
+  pub inviter_name: Option<String>,
+  pub status: AFWorkspaceInvitationStatus,
+  pub updated_at: DateTime<Utc>,
+}
+
+#[derive(Serialize, Deserialize, Eq, PartialEq, Debug, Clone)]
+#[repr(i16)]
```
appflowy, line 496: how about repr(u8) ?

tests/access_control/collab_ac_test.rs
```diff
@@ -0,0 +1,373 @@
...
+      )
+      .await?,
+    "stranger should not have write access"
+  );
+
+  Ok(())
+}
+
+#[sqlx::test(migrations = false)]
+async fn test_collab_access_control_send_receive_collab_update(pool: PgPool) -> anyhow::Result<()> {
```
appflowy, line 235: sometimes get pool timeout when using sqlx::test with pool

libs/database-entity/src/dto.rs
speed2exe: I have tried, smallest integer postgres can support is 16 bits

src/biz/workspace/ops.rs
speed2exe: ok, I'll add it in

speed2exe:
close in favor of https://github.com/AppFlowy-IO/AppFlowy-Cloud/pull/366 instead

PR #386, opened 2024-03-14
feat: Custom nginx ports

ThanatosDi:
Allow users to set the Nginx port themselves, defaulting to 80 and 443 if not set in the .env file.
Avoid ports conflicts, likely synology nas 80, 443 already used

### DIFF
--- a/deploy.env
+++ b/deploy.env
@@ -1,6 +1,10 @@
 # This file is a template for docker compose deployment
 # Copy this file to .env and change the values as needed
 
+# Nginx ports
+NGINX_HTTP_PORT=80
+NGINX_HTTPS_PORT=443
+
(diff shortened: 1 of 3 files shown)


appflowy:
@ThanatosDi Thank you for contributing. Will check it later

khorshuheng:
Since docker compose dev doesn't use nginx at all, we can omit the change for dev.env.

For port 443, in practice, most users are going to use an nginx instance outside docker compose for the purpose of supporting https traffic, so in long term, we will probably just keep port 80 within docker compose. But for now, it's fine to provide an option to map port 443.

khorshuheng:
> Looks like this is now available via #776, I think this PR can be closed now?

Yes, I will close this now.

PR #388, opened 2024-03-16
chore: integrate appflowy ai

appflowy:


suntorytimed:
@appflowy One question regarding this change. As there is a new env variable (`APPFLOWY_AI_URL`) added to the `docker-compose.yml`. Does this have any effect on a deployment environment or does this only affect dev environments deployed with the `docker-compose-dev.yml`? Are we expected to define an env variable `APPFLOWY_AI_URL` and `OPENAI_API_KEY` for production deployments?

appflowy:
Hi @suntorytimed. The AI feature is currently in the development stage and can only be accessed through the `docker-compose-dev.yml` file. We plan to extend its availability to the `docker-compose.yml` file in the future, making it accessible in production environments as well.

PR #426, opened 2024-03-29
feat: support client api to wasm

qinluhe:


Cargo.toml
```diff
@@ -143,6 +143,7 @@ members = [
     # services
     "services/collab-history",
     "services/realtime",
+    "libs/client-api-for-wasm"
```
appflowy, line 146: How about naming the crate `client-api-wasm`? This aligns with the naming conventions of other crates.

libs/client-api-for-wasm/tests/web.rs
```diff
@@ -0,0 +1,13 @@
...
+#![cfg(target_arch = "wasm32")]
+
+extern crate wasm_bindgen_test;
+use wasm_bindgen_test::*;
+
+wasm_bindgen_test_configure!(run_in_browser);
+
+#[wasm_bindgen_test]
+fn pass() {
+  assert_eq!(1 + 1, 2);
```
appflowy, line 12: should write true test?

libs/client-api/src/http.rs
```diff
@@ -278,17 +278,17 @@ impl Client {
...
-          // interact with Google services on behalf of the user without the user being actively
-          // engaged, access_type=offline is preferred to ensure long-term access.
-        .append_pair("access_type", "offline")
-          // In Google OAuth2.0, the prompt parameter is used to control the OAuth2.0 flow's behavior.
-          // It determines if the user is re-prompted for authentication and/or consent.
-          // 1. none: The authorization server does not display any authentication or consent user interface pages.
-          // 2. consent: The authorization server prompts the user for consent before returning information to the client
-          // 3. select_account: The authorization server prompts the user to select a user account.
-        .append_pair("prompt", "consent");
+				.query_pairs_mut()
```
appflowy, line 281: why these?

libs/client-api-for-wasm/src/entities.rs
```diff
@@ -0,0 +1,79 @@
+use client_api::error::ErrorCode;
+use serde::{Deserialize, Serialize};
+use tsify::Tsify;
+use wasm_bindgen::JsValue;
+
+#[derive(Tsify, Serialize, Deserialize, Default, Debug)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct Configuration {
```
appflowy, line 8: Can we reuse the structs including Config, ErrorCode etc. in client-api directly? then we don't need to define again.

.github/workflows/wasm_publish.yml
```diff
@@ -0,0 +1,61 @@
+name: Manual NPM Package Publish
+
+on:
```
appflowy, line 3: the ci should only trigger when creating a new tag on appflowy cloud

libs/client-api-for-wasm/src/logger.rs
```diff
@@ -0,0 +1,31 @@
+use crate::wasm_trace;
+
+pub struct WASMLogger;
+
+impl log::Log for WASMLogger {
```
appflowy, line 5: using `tracing-wasm` that can reuse all existing logs in the crate.( can take a look at appflowy_web/wasm-libs/af-wasm)

libs/client-api-for-wasm/src/entities.rs
qinluhe: Then we need to add
#[derive(Tsify, Serialize, Deserialize, Default, Debug)]
#[tsify(into_wasm_abi, from_wasm_abi)]
to these structs
I think it will make dirty

.github/workflows/wasm_publish.yml
qinluhe: This CI is manual, and it is not appropriate to bind the tag of this repo?

libs/client-api/src/http.rs
qinluhe: The format is automatically aligned when formatting

libs/client-api-for-wasm/src/entities.rs
appflowy: How about using cfg feature flag? Only enable when current target is wasm

PR #443, opened 2024-04-03
Test deploy token login

speed2exe:
- simplify magic link login
- fix magic login not working when another user is logged in


appflowy:
It's ok to merge if the wasm test fail.

PR #459, opened 2024-04-09
feat: expose magic link api

LucasXu0:


appflowy:
@LucasXu0 Can you add some tests for your changes. You can put it here
image

LucasXu0:
I added a simple test, but I think it's not sufficient. The sign-in process with a magic link involves the following steps:

1. The user inputs their email.
2. The system calls the sign-in with magic link API.
3. The user receives an email and then clicks the link in their mailbox.
4. The link launches AppFlowy.

However, the issue is that I don't know how to mock step 3, which is the most important part that should be covered.

libs/client-api/src/http.rs
```diff
@@ -215,6 +215,26 @@ impl Client {
     self.token.read().subscribe()
   }
 
+  /// Sign in with magic link
+  #[instrument(level = "debug", skip_all, err)]
+  pub async fn sign_in_with_magic_link(
+    &self,
+    email: &str,
+    redirect_to: Option<String>,
```
appflowy, line 223: Document how to use the redirect_to? what is the default redirect_to?

LucasXu0: Done

speed2exe:
@LucasXu0 @appflowy I think the test is sufficient. The link that is sent to the user will be same as `generate_sign_in_action_link`, which is already tested, so we don't need to do that.

PR #462, opened 2024-04-09
fix: handle receiver close for connection test

khorshuheng:
The real time connection test can still pass even though the the websocket client receive a 404 error while attempting to connect to the server, such as when LOCALHOST_WS environmental variable is set to the wrong path (eg. ws://localhost:8000`. In such scenario the connect future async loop returns early as the receiver error is not handled.


appflowy:
@khorshuheng  Thanks!

PR #463, opened 2024-04-11
chore: get collab http

speed2exe:


src/api/workspace.rs
```diff
@@ -86,6 +86,10 @@ pub fn workspace_scope() -> Scope {
         .route(web::put().to(update_collab_handler))
         .route(web::delete().to(delete_collab_handler)),
     )
+    .service(
+      web::resource("/v1/{workspace_id}/collab/{object_id}")
```
appflowy, line 90: maybe it need to update the workspace or collab access control middleware

PR #464, opened 2024-04-11
feat: support subscribe token

qinluhe:


libs/client-api-wasm/src/entities.rs
```diff
@@ -39,3 +50,106 @@ pub struct ClientResponse {
...
+  }
+}
+
+trait FromNumber {
+  fn from_number(n: i32) -> Self
+  where
+    Self: Sized;
+}
+
+impl FromNumber for CollabType {
```
appflowy, line 114: How about implement From<i32> for CollabType in Collab repo

PR #477, opened 2024-04-17
feat: using transaction to write multiple records

appflowy:
Spawns a background task that periodically writes cached collaboration data to disk in a single transaction. It consolidates multiple write operations into one transaction with the goal of minimizing database transaction pool usage,
thereby enhancing performance and reducing resource contention.

### DIFF
--- a/src/biz/collab/queue.rs
+++ b/src/biz/collab/queue.rs
@@ -1,0 +1,391 @@
+use crate::biz::collab::cache::CollabCache;
+use crate::biz::collab::queue_redis_ops::{
+  get_pending_meta, get_remaining_pending_write, insert_pending_meta, remove_pending_meta,
+  storage_cache_key,
+};
+
+use crate::state::RedisConnectionManager;
+use anyhow::Context;
+use app_error::AppError;
+use collab_entity::CollabType;
+use database_entity::dto::{CollabParams, QueryCollab, QueryCollabResult};
+use serde::{Deserialize, Serialize};
+
+use std::cmp::Ordering;
+use std::collections::{BinaryHeap, HashMap};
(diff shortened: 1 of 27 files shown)


src/biz/collab/queue.rs
```diff
@@ -0,0 +1,467 @@
...
+  drop(pending_lock);
+
+  write_items
+    .iter()
+    .map(|pending| storage_cache_key(&pending.object_id, pending.data_len))
+    .collect()
+}
+
+#[derive(Clone)]
+pub(crate) struct PendingWrite {
```
speed2exe, line 394: I supposed this is stored in mem. For scalability, I would suggest using redis sorted set

src/biz/collab/queue.rs
```diff
@@ -0,0 +1,467 @@
...
+  /// Enqueues a object for deferred processing. High priority writes are processed before low priority writes.
+  ///
+  /// adds a write task to a pending queue, which is periodically flushed by another task that batches
+  /// and writes the queued collaboration objects to a PostgreSQL database.
+  ///
+  /// This data is stored temporarily in the `collab_cache` and is intended for later persistent storage
+  /// in the database. It can also be retrieved during subsequent calls in the [CollabStorageImpl::get_encode_collab]
+  /// to enhance performance and reduce database reads.
+  ///
+  pub async fn push(
```
speed2exe, line 80: if possible, should check current load and flush to postgres if exceeds certain limit, this will mitigate scenario with sudden spike.

src/biz/collab/queue.rs
appflowy: done

Issue #411, opened 2024-03-23
[FR] Separate WebSocket Service from AppFlowy Cloud Server for Improved Stability and Stateless Operation


appflowy:
## Background
We are currently using actix-web to implement our WebSocket service within the AppFlowy cloud server infrastructure. This integration means that whenever the AppFlowy cloud server is restarted, the WebSocket service is also restarted. This restart leads to the loss of all existing WebSocket connections. To mitigate this, we have implemented auto-reconnect logic on the client side, which allows clients to automatically reconnect to the WebSocket server.

However, for long-term stability and to move towards a stateless architecture for the AppFlowy cloud server, we propose deploying the WebSocket service as a separate, standalone service. This separation would have several benefits:

1. **Improved Stability**: By decoupling the WebSocket service from the main AppFlowy cloud server, WebSocket connections would not be dropped due to the restarts of the cloud server, leading to a more stable connection experience for users.

2. **Stateless Cloud Server**: Separating the WebSocket service would be a step towards making the AppFlowy cloud server stateless, enhancing scalability and simplifying maintenance and updates.

3. **Focused Scaling and Management**: With the WebSocket service operating independently, it could be scaled based on connection load independently of the AppFlowy cloud server, allowing for more efficient resource use and focused troubleshooting and performance tuning.

## What to do
1. moving our WebSocket code from `src/biz/actix_ws` to `services/realtime`.
2. deploying the WebSocket service using Docker, making it easier to manage, scale, and isolate from other parts of our system.
3. support docker-compose-dev.yml and docker-compose.yml.

khorshuheng:
Based on my current understanding, the websocket service API and the cloud server API share most of the code under `biz` crate. As such, if we are to move the websocket code to `services/realtime`, it's likely that `biz/collab` , `biz/snapshot` , `biz/user`, `biz/workspace` will all have to become independent libraries (or become part of the existing local libraries).

The other alternative which i can think of, is to keep the websocket code within the current directory, but build two different binaries, one for the websocket service and the other for the server API.

We could also add a command line argument / env config, which will start either a websocket service or a cloud server, depending on the argument / config value.

Which of the above options are preferable?

appflowy:
I'm reviewing the `biz/actix_ws` module, which mainly depends on the `collab_rt` crate. Considering this dependency, it seems possible to move it to `services/realtime`. This transition would require adjustments in `api/workspace.rs` and `application.rs` to adapt to the new structure.

[...]

PR #483, opened 2024-04-20
Draft: feat: Add realtime service as proxy to websocket server

khorshuheng:
This is the draft PR based on the version 2 design as proposed in https://github.com/AppFlowy-IO/AppFlowy-Cloud/issues/411#issuecomment-2030966504 .

As of the moment the realtime service is purely acting as a proxy without any connection retry like https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/libs/client-api/src/native/retry.rs#L107 and https://github.com/AppFlowy-IO/AppFlowy/blob/9b4fd3bf57cf27dff70592233c9b45430a7a00a1/frontend/rust-lib/flowy-server/src/af_cloud/server.rs#L244 , so the connection to the client will still be disrupted in the event of appflowy server restart.

Even with retry added on the real time service, the stability is probably not much better than client side retry, though it does help that we can configure the retry configuration anytime via the real time service as opposed to updating every client.

Version 3 should be the long term solution, though there doesn't appear to be much code that we can reuse from refactoring version 2 to version 3, since the actor design will be very different.

I will be looking into the collab-rt server to see how much effort will be needed if we jump straight to version 3.

Other items that are not implemented yet beside connection retry:
- This does not handle the scenario of larger websocket payload, which is not handled via the ws/v1 endpoint.
- Haven't properly handled all message delivery failure scenario yet

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4545,6 +4545,24 @@ dependencies = [
 [[package]]
 name = "realtime"
 version = "0.1.0"
+dependencies = [
+ "actix",
+ "actix-http",
+ "actix-rt",
+ "actix-web",
+ "actix-web-actors",
+ "anyhow",
+ "app-error",
+ "async-trait",
+ "bytes",
+ "client-websocket",
+ "dotenvy",
+ "futures-util",
+ "thiserror",
+ "tokio",
+ "tokio-tungstenite",
+ "tracing",
+]
 
 [[package]]
 name = "redis"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,11 +6,11 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-actix = "0.13.3"
-actix-web = { version = "4.5.1", default-features = false, features = ["openssl", "compress-brotli", "compress-gzip"] }
+actix.workspace = true
+actix-web = { workspace = true, default-features = false, features = ["openssl", "compress-brotli", "compress-gzip"] }
 actix-http = { workspace = true, default-features = false, features = ["openssl", "compress-brotli", "compress-gzip"] }
-actix-rt = "2.9.0"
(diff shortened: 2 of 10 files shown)


khorshuheng:
Originally, i was also thinking if i can move the authentication to real time service, and interact with the RealtimeServer actor directly. But i realized that this will:
1. Introduce an endpoint on Appflowy server that is accessible without authentication
2. Might require significant changes to RealtimeServer actor / an entirely new implementation of the actor

Which is the reason the current draft has very little business logic in it beyond passing the message to appflowy server.

appflowy:
> Even with retry added on the real time service, the stability is probably not much better than client side retry, though it does help that we can configure the retry configuration anytime via the real time service as opposed to updating every client.
Yes, allowing the server to control retry configurations is more efficient. It can also dynamically adjust settings based on the current network load

>I will be looking into the collab-rt server to see how much effort will be needed if we jump straight to version 3.
The collab-rt server is designed for horizontal scaling and operates as a unified service. Feel free to reach out if you need assistance or want to discuss how to effectively deploy it as a single service while upgrading to version 3.

appflowy:
> Originally, i was also thinking if i can move the authentication to real time service, and interact with the RealtimeServer actor directly. But i realized that this will:
> 
> 1. Introduce an endpoint on Appflowy server that is accessible without authentication

Do you mean we could use HTTP requests or gRPC to communicate with the Appflowy server? This way, the realtime server wouldn't need to manage authentication directly.

> 2. Might require significant changes to RealtimeServer actor / an entirely new implementation of the actor
> 
> Which is the reason the current draft has very little business logic in it beyond passing the message to appflowy server.

Got it

appflowy:
@khorshuheng FYI. I am going to move the collab-rt crate to services/collab-rt

khorshuheng:
Closed in favor of https://github.com/AppFlowy-IO/AppFlowy-Cloud/pull/502 .

PR #514, opened 2024-05-01
feat: add more detailed logging around errors in Transaction::try_encode_state_as_update_v1

Horusiath:
This would allow us to attach internal document state to errors, that cause panics during applying CRDT state updates.

I suspect that the reason behind these panics was fixed in https://github.com/y-crdt/y-crdt/pull/405 but we won't be able to confirm that until we upgrade to the latest version or get the log data.


appflowy:
@Horusiath Understood. Before we upgrade to version 0.18.2, we can use this method to verify the issue.

PR #525, opened 2024-05-05
fix: make sure test for offline user connect then sync failed when one or more future task failed

khorshuheng:
The test, `simulate_50_offline_user_connect_and_then_sync_document_test` , currently passed even though one or more task failed due to sync failures. Added a check to make sure that all the task completed successfully.


khorshuheng:
There's also a similar test below that was commented out, and the commented test does check if all the task succeeded. Should the commented test be kept or removed?

appflowy:
@khorshuheng The commented tests will be enabled in the future. For now, they are flaky tests

lilpoozie2005 (approved):

PR #526, opened 2024-05-06
feat: add get user workspace

qinluhe:


libs/client-api/src/http.rs
```diff
@@ -963,6 +963,30 @@ impl Client {
     AppResponse::<()>::from_response(resp).await?.into_error()
   }
 
+  // Support browser, because browser does not support GET method with body
+  #[instrument(level = "info", skip_all, err)]
+  pub async fn batch_post_collab(
```
appflowy, line 968: the only difference between batch_get_collab and batch_post_collab is the http method?

qinluhe: yes

qinluhe: > the only difference between batch_get_collab and batch_post_collab is the http method?

yes

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #531, opened 2024-05-07
[WIP] feat: add postgres vector extension for semantic search indexing

Horusiath:
**It's just a draft**: We didn't decide if we're going this route, but it's a proof-of-concept to decide on operational complexity of using semantic search directly inside of PostgreSQL.


appflowy:
seems duplicate with https://github.com/AppFlowy-IO/AppFlowy-Cloud/pull/560.

PR #559, opened 2024-05-17
chore: integrate redis stream

appflowy:


services/appflowy-collaborate/src/group/broadcast.rs
```diff
@@ -105,6 +115,14 @@ impl CollabBroadcast {
             event.update.len(),
             origin
           );
+
+          let stream_update = event.update.clone();
+          let cloned_update_streaming = update_streaming.clone();
+          // FIXME: how to maintain the order of the update when using spawn here.
+          tokio::spawn(async move {
+            cloned_update_streaming.send_update(stream_update).await;
```
Horusiath, line 123: Instead of using spawn, use a queue ie. `tokio::sync::mpsc::UnboundedSender` and send the updates through it. Then spawn a tokio task which will consume incoming updates from `UnboundedReceiver` on the other end.

Horusiath (changes requested):
One note regarding FIXME comment.

services/appflowy-collaborate/src/group/group_init.rs
```diff
@@ -285,6 +300,89 @@ impl EditState {
...
+  async fn send_update(&self, update: Vec<u8>) {
+    if let Some(redis_update_stream) = self.update_stream.lock().await.deref_mut() {
+      if let Err(err) = redis_update_stream
+        .insert_message(CollabUpdateEvent::UpdateV1 {
+          encode_update: update,
+        })
+        .await
+      {
+        //FIXME: If fail to insert update to redis stream, which will cause missing update issue.
+        error!(
```
Horusiath, line 352: Just like described above: if you use queues, you can make queue consumer task inside of collab update streaming. That task can return a `Result` which will propagate the error to `CollabUpdateStreaming` controller object (similar to how HTTP `server.await` works in main).

I propose this approach since the `CollabUpdateStreaming` has no control over itself, so it's not able to properly  restart on its own. Therefore critical failures needs to be propagated upwards.

services/appflowy-collaborate/src/group/group_init.rs
```diff
@@ -285,6 +300,89 @@ impl EditState {
...
+    .into_iter()
+    .flat_map(|update| {
+      StreamBinary::try_from(CollabUpdateEvent::UpdateV1 {
+        encode_update: update,
+      })
+      .ok()
+    })
+    .collect::<Vec<StreamBinary>>();
+  // FIXME: How to handle the situation when there are too many messages that need to be sent
+  if let Err(err) = redis_update_stream.insert_messages(messages).await {
```
Horusiath, line 381: Yrs has method called [merge_updates](https://docs.rs/yrs/latest/yrs/fn.merge_updates_v1.html), so you could squash multiple updates into one).

Horusiath (changes requested):

services/appflowy-collaborate/src/group/broadcast.rs
appflowy: Agreed. I used this approach in client-api. It will merge multiple updates within a period of time.

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #560, opened 2024-05-17
ai pgvector

speed2exe:
- added pg for appflowy ai to all `docker-compose.yml`and `env` files


migrations/20240507041927_vector-ext.sql
```diff
@@ -0,0 +1,2 @@
+-- Create the 'vector' extension within the database that is set in the docker-compose.yml
+CREATE EXTENSION IF NOT EXISTS vector;
```
appflowy, line 2: Is this migration only need to run on appflowy-ai postgres DB?

speed2exe: yes i believe so.

appflowy: But how to run migration for appflowy-ai postgres?  IIRC, Migrations under migrations folder only work for appflowy-cloud postgres database.

speed2exe: I dont think there's no migration required. appflowy ai will only need a pg db with extension install. @Horusiath correct me if I'm wrong

hicasper: > I dont think there's no migration required. appflowy ai will only need a pg db with extension install. @Horusiath correct me if I'm wrong

try https://hub.docker.com/r/tensorchord/pgvecto-rs , there is no need to build  postgresql image with extension again, or you can build image and upload to docker hub (since not everyone has enough space and time to run the build).

PR #576, opened 2024-05-26
chore: share library for authentication

khorshuheng:
Authentication related functionality is extracted into a shared library to enable sharing across different services.

Additional changes:
- spawn_blocking_with_tracing is moved from the telemetry module to the authentication crate, as this function is used only within authentication.


libs/authentication/src/jwt.rs
```diff
@@ -99,14 +97,8 @@ impl FromRequest for Authorization {
...
 
-// impl RealtimeUser for Authorization {
-//   fn id(&self) -> &str {
-//     &self.uuid_str
-//   }
-// }
-
 fn get_auth_from_request(req: &HttpRequest) -> Result<Authorization, actix_web::Error> {
-  let state = req.app_data::<Data<AppState>>().unwrap();
+  let jwt_secret_data = req.app_data::<Data<Secret<String>>>().unwrap();
```
appflowy, line 101: It needs to return an error instead of unwrap

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #608, opened 2024-06-06
let prometheus track open ai tokens used on search

Horusiath:
This PR registers a new `search_tokens_used` counter in prometheus so that we can track how many OpenAI API tokens are used over time.


speed2exe:
@Horusiath are the AI token usage tied to workspace or user?

Horusiath:
Right now they are not grouped by anything. After talking with @appflowy haven't decided yet on what attribute group them by.

speed2exe (approved):
LGTM. can merge if there's no more changes

speed2exe:
@Horusiath No problems, we can add them later if needed.

Horusiath:
I've added workspace id as a label to token metrics (as we discussed it with @appflowy ). With that state of the PR, we are ready to merge it.

lilpoozie2005 (approved):

PR #618, opened 2024-06-13
chore: move billing client

speed2exe:


libs/billing/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "billing"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+client-api = { path = "../client-api" }
+shared-entity = { path = "../shared-entity" }
```
appflowy, line 8: can use shared-entity.workspace = true

lilpoozie2005 (approved):

PR #625, opened 2024-06-15
feat: support uploading file with multiple part

appflowy:
1. Support multipart upload functionality and update the client API accordingly.
2. Remove all objects for a given workspace when that workspace is deleted. Previously, the old approach removed blob metadata and S3 objects one by one. Now, simply remove the specific directory corresponding to the workspace ID.
3. Replace the s3-rust crate with the aws-sdk-s3 crate, as s3-rust does not support multipart upload.
4. Add more tests


src/application.rs
```diff
@@ -140,7 +146,10 @@ pub async fn run_actix_server(
       // .wrap(DecryptPayloadMiddleware)
       .wrap(access_control.clone())
       .wrap(RequestIdMiddleware)
-      .app_data(web::JsonConfig::default().limit(5 * 1024 * 1024))
+      .app_data(web::JsonConfig::default().limit(10 * 1024 * 1024).error_handler(|err, req| {
```
speed2exe, line 149: maybe can remove this part, since the size limitation is not enforced here

src/api/file_storage.rs
```diff
@@ -42,17 +49,114 @@ pub fn file_storage_scope() -> Scope {
       web::resource("/{workspace_id}/blobs")
         .route(web::get().to(get_all_workspace_blob_metadata_handler)),
     )
+    .service(web::resource("/{workspace_id}/create_upload").route(web::post().to(create_upload)))
```
speed2exe, line 52: there's `workspace_id` in the path, it should be enforced and used as prefix for object key in the server, else this allow client to store in object using any key, potentially affecting other user

appflowy: It creates a upload id. The object_key will be used in the upload part api endpoint

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #630, opened 2024-06-18
chore: change open ai token tracking to daily basis

Horusiath:


services/appflowy-indexer/src/collab_handle.rs
```diff
@@ -363,13 +363,14 @@ mod test {
...
-        .unwrap()
-        .get(0);
+    let tokens: i64 = sqlx::query(
+      "SELECT index_tokens_consumed from af_workspace_ai_usage WHERE workspace_id = $1",
+    )
+    .bind(workspace_id)
+    .fetch_one(&db)
+    .await
+    .unwrap()
+    .get(0);
```
speed2exe, line 373: @Horusiath 
maybe can consider to use `query_scalar!` instead
the `unwrap()` can also change to `?`

lilpoozie2005 (approved):

PR #633, opened 2024-06-19
fix: s3 config when deploy without minio

appflowy:


speed2exe:
LGTM

src/application.rs
```diff
@@ -382,7 +384,7 @@ async fn get_redis_client(redis_uri: &str) -> Result<redis::aio::ConnectionManag
 }
 
 pub async fn get_aws_s3_client(s3_setting: &S3Setting) -> Result<aws_sdk_s3::Client, Error> {
-  trace!("Connecting to S3 with setting: {:?}", &s3_setting);
+  info!("Connecting to S3 with setting: {:?}", &s3_setting);
```
Horusiath, line 387: Arguably S3 `access_key` should also be hidden behind secret, but it's not a blocking issue.

appflowy: I will remove this print

lilpoozie2005 (approved):

PR #634, opened 2024-06-20
chore: fix clippy linting errors and warnings

khorshuheng:
This PR fixed the errors reported by Clippy, which includes:
- unused struct fields
- reference that is immediately dereferenced by the compiler


khorshuheng:
Close as integration test will fail for PR originating from forked repository

PR #636, opened 2024-06-20
chore: allow collab persistence and edit state behaviour to be configurable

khorshuheng:
The following parameter values are currently hard coded:
- Maximum allowable edit count and duration
- Interval for persisting in-memory collab group to cache/disk

Allowing the above parameters to be configured will allow us to control how frequent the in-memory edit states are saved to the cache.


khorshuheng:
Closing this PR as the integration workflow cannot be run successfully from a PR originating from a fork.

PR #642, opened 2024-06-22
feat: Integrate workspace setting

appflowy:


libs/database-entity/src/dto.rs
```diff
@@ -511,10 +511,47 @@ pub struct AFWorkspace {
 #[derive(Serialize, Deserialize)]
 pub struct AFWorkspaces(pub Vec<AFWorkspace>);
 
-#[derive(Default, Serialize, Deserialize)]
+#[derive(Serialize, Deserialize)]
 pub struct AFWorkspaceSettings {
   #[serde(default)]
-  pub disable_indexing: bool,
+  pub disable_search_indexing: bool,
```
Horusiath, line 517: This will break related Postgres queries.

lilpoozie2005 (approved):

PR #645, opened 2024-06-24
Merge indexer collaborate

Horusiath:


services/appflowy-collaborate/src/collab/disk_cache.rs
```diff
@@ -54,6 +56,17 @@ impl CollabDiskCache {
     transaction: &mut Transaction<'_, sqlx::Postgres>,
   ) -> AppResult<()> {
     insert_into_af_collab(transaction, uid, workspace_id, params).await?;
+    if let Some(em) = &params.embeddings {
```
appflowy, line 59: Is this intend to log these inforamtion? it might be lots of logs

Horusiath: Tbh. I don't think it will be that much. Are we keeping the metrics on how often collabs are inserted into database?

appflowy: No. we don't implement the metrics yet. Will see how it going

libs/database/src/collab/collab_db_ops.rs
```diff
@@ -76,11 +76,11 @@ pub async fn insert_into_af_collab(
       if existing_workspace_id == workspace_id {
         sqlx::query!(
           "UPDATE af_collab \
-        SET blob = $2, len = $3, partition_key = $4, encrypt = $5, owner_uid = $6 WHERE oid = $1",
```
Horusiath, line 79: We cannot update `partition_key` as it's a part of the primary key on `af_collab` table. This would cause all foreign keys to crash.

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #652, opened 2024-06-26
chore: stop indexing when stream is not exist

appflowy:


Horusiath (approved):
Looks good. The only issue is that so far we manager to encounter this issue only on test env (never on localhost), so we probably won't be able to confirm it without deploying it on test.

lilpoozie2005 (approved):

PR #687, opened 2024-07-12
feat: support protobuf based serialization for collab params

khorshuheng:
Allow CollabParams to be deserialized using both protobuf and bincode encoding. Also add a method where the client can choose to serialized CollabParams using protobuf instead of bincode.


khorshuheng:
CollabParams is currently used as the payload from the client to the server during batch collab creation / data import. The struct is encoded before being sent by the http client, and decoded on the cloud server.

In order to support older clients, which will continue to send bincode as payload, the server will check the request header for the presence of `X-Serialization-Type` . If not specified, it's assumed that the request comes from an old client, and the payload is of the bincode cormat. Otherwise, if `X-Serialization-Type` is `protobuf` , then the server will treat the payload as protobuf and deserialize it accordingly.

A new endpoint, `/api/server_info` has also been introduced, so that the newer clients can use this endpoint as a mean to check whether the payload should be sent as bincode or protobuf. If calls to `/api/server_info` returns 404, it's assumed that the server is not new enough, and hence incapable of handling protobuf payload. Bincode payload will be sent instead. Otherwise, the payload will be sent as protobuf.

khorshuheng:
The other alternatives which have been considered:
- The newer clients will call the current API endpoint using protobuf as payload. If the server return bad request, the clients will send bincode as a fallback.
- A new API endpoint will be introduced, that will deserialize the payload in protobuf. If the client is unable to reach this endpoint, the old API endpoint is called instead as a fallback.

The alternatives above has issues when the payload is large, such as the scenario tested in `batch_insert_collab_success_test` . Currently, our server returns the response early if deserialization error is encountered, even though the payload stream has not been completely consumed. This appears to cause issues with the connection between the reqwest client and the server, in which subsequent request will timeout indefinitely, unless a new connection is created, or sufficient time has passed. 

If we consumed all the payload before returning the response code, regardless of serialization error, then the alternatives proposed above are viable. However, since such fix cannot be retroactively applied to old server versions, it still won't help us to make newer clients to be compatible with older servers.

libs/client-api/src/http.rs
```diff
@@ -844,9 +847,12 @@ impl Client {
     &self,
     method: Method,
     url: &str,
+    serialization_type: &SerializationType,
   ) -> Result<RequestBuilder, AppResponseError> {
     #[cfg(feature = "enable_brotli")]
     {
+      let serialization_type_header_value =
+        HeaderValue::from_str(serialization_type.to_string().as_str()).unwrap();
```
appflowy, line 855: is it safe to unwrap?

khorshuheng: The only scenario where this will panic is if we wrongly include non-ASCII characters like `\n` in the to string method for Serialization Type enum. Hence, this should be safe to unwrap.

appflowy: Understood. It would be better to make a annotation for this case. For example: //safety: xxx

appflowy:
@khorshuheng  Will we have another PR for replacing  `RealtimeMessage` with [realtime.proto](https://github.com/AppFlowy-IO/AppFlowy-Collab/blob/main/collab-entity/proto/collab/realtime.proto)?

khorshuheng:
> @khorshuheng Will we have another PR for replacing `RealtimeMessage` with [realtime.proto](https://github.com/AppFlowy-IO/AppFlowy-Collab/blob/main/collab-entity/proto/collab/realtime.proto)?

Yes, but not that soon. RealtimeMessage is a lot harder to migrate due to it being used almost everywhere. Will need to carefully assess the impact.

appflowy:
> > @khorshuheng Will we have another PR for replacing `RealtimeMessage` with [realtime.proto](https://github.com/AppFlowy-IO/AppFlowy-Collab/blob/main/collab-entity/proto/collab/realtime.proto)?
> 
> Yes, but not that soon. RealtimeMessage is a lot harder to migrate due to it being used almost everywhere. Will need to carefully assess the impact.

Ok. This PR seems ready to merge. I will test it today and then merge it

PR #690, opened 2024-07-15
chore: remove pg advisory lock

Horusiath:
We noticed that occasionally Postgres advisory lock created when creating a new user is grinding the database down them down to halt: since it's an entire table lock rather than row-level lock. 

This PR is supposed to help by removing the advisory lock. Additionally concurrent conflicts caused by unlikely situation when two requests will try to create the same user will result in HTTP error response for one of them.


appflowy:
The Advisory locks can be acquired at session level or at transaction level. it seems like we use it on transaction level. it shouldn't lock entire table.

PR #698, opened 2024-07-23
feat: Add create, retrieve, delete API for publish view global comment

khorshuheng:
Add API endpoints for creating, retrieving and deleting global comments from a published view. Most features are ready, apart from avatar url, which i tentatively set as None for all cases.


migrations/20240723090305_publish_view_comment.sql
```diff
@@ -0,0 +1,19 @@
+-- stores the comments on a published view
+CREATE TABLE IF NOT EXISTS af_published_view_comment (
+  comment_id          UUID NOT NULL DEFAULT gen_random_uuid(),
+  -- comments are never deleted, only marked as deleted, unless we intentionally wants to clean
+  -- the tables by removing the comments from the database
+  reply_comment_id    UUID REFERENCES af_published_view_comment(comment_id) ON DELETE CASCADE,
+  -- The view id should exists on af_published_collab, However, we can't enforce this foreign key
+  -- constraint because af_published_collab primary key is (workspace_id, view_id)
```
speed2exe, line 8: ideally we should delete the comments when the it's unpublished, not sure if adding `workspace_id` column can help with this, or might have to add a postgres trigger to achieve. if not deleted, then if user publish a doc, comments, unpublished and publish the same doc again, the comments is probably still there

speed2exe:
@khorshuheng left a comment, the rest seems fine. will user reactions to comments in another PR or will be added to this?

migrations/20240723090305_publish_view_comment.sql
khorshuheng: According to the requirement, if it is the same view id, the comments should still be preserved, similar to how a blog post page still keep the comments despite update.

khorshuheng:
> @khorshuheng left a comment, the rest seems fine. will user reactions to comments in another PR or will be added to this?

User reaction will be  in another PR.

migrations/20240723090305_publish_view_comment.sql
khorshuheng: That said, the lack of constraint is bothering me as well - because that means it is possible for comment to be posted on non existent view id.

qinluhe (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #700, opened 2024-07-24
Duplicate published doc

speed2exe:


libs/client-api/src/http.rs
```diff
@@ -652,6 +653,23 @@ impl Client {
       .into_data()
   }
 
+  #[instrument(level = "info", skip_all, err)]
+  pub async fn get_workspace_folder(
```
appflowy, line 657: can we add documentation for this api endpoint?

src/biz/workspace/publish_dup.rs
```diff
@@ -0,0 +1,941 @@
...
+      add_to_view_info(acc, child_views);
+    }
+  }
+}
+
+fn collab_from_doc_state(doc_state: Vec<u8>, object_id: &str) -> Result<Collab, AppError> {
+  let collab = Collab::new_with_source(
+    CollabOrigin::Server,
+    object_id,
+    DataSource::DocStateV1(doc_state),
```
appflowy, line 921: should reference to the version in `EncodeCollab`
image

libs/database/src/publish.rs
```diff
@@ -0,0 +1,272 @@
...
+
+  Ok(res)
+}
+
+#[inline]
+pub async fn insert_or_replace_publish_collabs<'a, E: Executor<'a, Database = Postgres>>(
+  executor: E,
+  workspace_id: &Uuid,
+  publisher_uuid: &Uuid,
+  publish_item: &[PublishCollabItem<serde_json::Value, Vec<u8>>],
```
appflowy, line 109: should use Vec<PublishCollabItem> then you don't need to clone

libs/database/src/publish.rs
```diff
@@ -0,0 +1,272 @@
...
+  let publish_names: Vec<String> = publish_item
+    .iter()
+    .map(|item| item.meta.publish_name.clone())
+    .collect();
+  let metadatas: Vec<serde_json::Value> = publish_item
+    .iter()
+    .map(|item| item.meta.metadata.clone())
+    .collect();
+
+  let blobs: Vec<Vec<u8>> = publish_item.iter().map(|item| item.data.clone()).collect();
```
appflowy, line 121: no need to clone

libs/shared-entity/src/dto/workspace_dto.rs
```diff
@@ -121,7 +122,165 @@ pub struct CollabResponse {
...
+}
+
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
+pub struct FolderView {
+  pub view_id: String,
+  pub name: String,
+  pub icon: Option<ViewIcon>,
+  pub is_space: bool,
+  pub is_private: bool,
+  pub extra: Option<serde_json::Value>,
```
appflowy, line 138: what is the extra used for? it would be better to have documentation here

libs/shared-entity/src/dto/workspace_dto.rs
```diff
@@ -1,8 +1,9 @@
 use chrono::{DateTime, Utc};
 use collab_entity::{CollabType, EncodedCollab};
+use collab_folder::{Folder, ViewIcon};
```
appflowy, line 3: shouldn't reference to `collab_folder` here.

libs/shared-entity/Cargo.toml
```diff
@@ -28,6 +28,8 @@ validator = { version = "0.16", features = ["validator_derive", "derive"], optio
 futures = "0.3.30"
 bytes = "1.6.0"
 log = "0.4.21"
+collab-folder = { workspace = true }
```
appflowy, line 31: can we remove collab-folder here?

services/appflowy-collaborate/src/group/broadcast.rs
```diff
@@ -216,10 +216,13 @@ impl CollabBroadcast {
 
                   trace!("[realtime]: send {} => {}", message, cloned_user.user_device());
                   if let Err(err) = sink.send(message).await {
-                    error!("fail to broadcast message:{}", err);
+                    tracing::error!("fail to broadcast message:{}", err);
```
appflowy, line 219: it's not necessary to add `tracing::` here

services/appflowy-collaborate/src/group/broadcast.rs
```diff
@@ -314,7 +317,7 @@ async fn handle_client_messages<Sink>(
         Ok(response) => {
           trace!("[realtime]: sending response: {}", response);
           match sink.send(response.into()).await {
-            Ok(_) => {},
+            Ok(()) => {},
```
appflowy, line 320: what is the difference using `()` instead of `_`

services/appflowy-collaborate/src/group/mod.rs
```diff
@@ -1,7 +1,7 @@
 pub(crate) mod broadcast;
 pub(crate) mod cmd;
 pub(crate) mod group_init;
-pub(crate) mod manager;
+pub mod manager;
```
appflowy, line 4: it seems pub(crate is enough.

services/appflowy-collaborate/src/lib.rs
```diff
@@ -8,7 +8,7 @@ pub mod compression;
 pub mod config;
 pub mod connect_state;
 pub mod error;
-mod group;
+pub mod group;
```
appflowy, line 11: it should be private too

services/appflowy-collaborate/src/metrics.rs
```diff
@@ -84,7 +84,7 @@ impl CollabRealtimeMetrics {
 }
 
 #[derive(Clone, Default)]
-pub(crate) struct CollabMetricsCalculate {
+pub struct CollabMetricsCalculate {
```
appflowy, line 87: should be private

src/biz/collab/ops.rs
```diff
@@ -137,3 +151,79 @@ pub async fn get_collab_member_list(
...
+    uid,
+    workspace_id,
+    workspace_id,
+    CollabType::Folder,
+  )
+  .await?;
+  let folder = Folder::from_collab_doc_state(
+    uid,
+    CollabOrigin::Server,
+    DataSource::DocStateV1(encoded_collab.doc_state.to_vec()),
```
appflowy, line 192: please review all the places that using encode_collab. You can use into() instead
image

libs/client-api/src/http.rs
speed2exe: added

src/biz/workspace/publish_dup.rs
speed2exe: dont this is possible, as the published blobs are `doc_state` (Bytes) that is assumed to have v1, instead of bytes of `EncodedCollab`

libs/database/src/publish.rs
speed2exe: changed

libs/database/src/publish.rs
speed2exe: removed the clone

libs/shared-entity/src/dto/workspace_dto.rs
speed2exe: will add in some documentation here. AFAIK it currently holds information such as if the view is a space and font information

libs/shared-entity/Cargo.toml
speed2exe: this is required and used (`Folder`, `ViewIcon`) for the returned folder structure (`FolderView`)

services/appflowy-collaborate/src/group/broadcast.rs
speed2exe: i prefer `()` as it gives move information(that the returned value is nothing, so we dont have to anything), as compared to `_` which indicate that the returned value can be something, but we choose to ignore it

src/biz/collab/ops.rs
speed2exe: done

services/appflowy-collaborate/src/group/broadcast.rs
speed2exe: removed

libs/shared-entity/src/dto/workspace_dto.rs
speed2exe: I can move the logic of building the folder view. But for `ViewIcon`(required for web display), I'll have to duplicate from the `collab-folder`.

```rust
#[derive(Eq, PartialEq, Debug, Hash, Clone, Serialize_repr, Deserialize_repr)]
#[repr(u8)]
pub enum IconType {
  Emoji = 0,
  Url = 1,
  Icon = 2,
}

#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
pub struct ViewIcon {
  pub ty: IconType,
  pub value: String,
}
```

libs/shared-entity/Cargo.toml
speed2exe: this also required to represents the published data in the frontend, which the structs can then be reused

libs/shared-entity/src/dto/workspace_dto.rs
appflowy: move `collab_folder_to_folder_view` to `get_user_workspace_structure` then you don't need to depend on collab-folder

libs/shared-entity/Cargo.toml
appflowy: move collab_folder_to_folder_view to get_user_workspace_structure then you don't need to depend on collab-folder

speed2exe: removed `collab_folder` dep on `shared-entitiy`

libs/shared-entity/src/dto/workspace_dto.rs
speed2exe: removed `collab_folder` dep on `shared-entitiy`

services/appflowy-collaborate/src/lib.rs
speed2exe: this wont if it's private

services/appflowy-collaborate/src/group/mod.rs
speed2exe: it wont compile if I use `pub (crate)`

services/appflowy-collaborate/src/metrics.rs
speed2exe: does not compile if it's only public at crate level

libs/database/src/publish.rs
```diff
@@ -0,0 +1,273 @@
...
+    &metadatas,
+    &blobs,
+    item_count as i32,
+  )
+  .execute(executor)
+  .await?;
+
+  if res.rows_affected() != item_count as u64 {
+    tracing::warn!(
+      "Failed to insert or replace publish collab meta batch, workspace_id: {}, publisher_uuid: {}, rows_affected: {}",
```
Horusiath, line 153: This is a partial failure - some rows might have been updated, some might not:

1. If you want to be strongly consistent (either all collabs were duplicated or none of them are), you should rollback transaction at this point.
2. If you want to be elastic, we should probably return row_affected count and inform user how many collabs have we duplicated.

Also log both `{rows_affected}` and `{item_count}` so that we have a clue how many were untouched.

libs/database/src/publish.rs
```diff
@@ -0,0 +1,273 @@
...
+    workspace_id,
+  )
+  .fetch_one(executor)
+  .await?;
+
+  Ok(res)
+}
+
+#[inline]
+pub async fn insert_or_replace_publish_collabs<'a, E: Executor<'a, Database = Postgres>>(
```
Horusiath, line 105: Generally speaking this might be a very heavy query. We know that individual collabs can have MBs in size. If there's no upper bound on their size/count, we might have as well lock the table for undefined time.

libs/database/src/publish.rs
```diff
@@ -0,0 +1,273 @@
...
+
+pub async fn select_published_collab_info<'a, E: Executor<'a, Database = Postgres>>(
+  executor: E,
+  view_id: &Uuid,
+) -> Result<PublishInfo, AppError> {
+  let res = sqlx::query_as!(
+    PublishInfo,
+    r#"
+      SELECT
+        (SELECT publish_namespace FROM af_workspace aw WHERE aw.workspace_id = apc.workspace_id) AS namespace,
```
Horusiath, line 261: This could be LEFT JOIN ;)

services/appflowy-collaborate/src/group/cmd.rs
```diff
@@ -169,6 +182,40 @@ where
...
+    messages: Vec<ClientCollabMessage>,
+  ) -> Result<(), RealtimeError> {
+    if messages.is_empty() {
+      warn!("Unexpected empty collab messages sent from server");
+      return Ok(());
+    }
+
+    if let Some(group) = self.group_manager.get_group(&object_id).await {
+      let (collab_message_sender, _collab_message_receiver) = futures::channel::mpsc::channel(1);
+      let (mut message_by_oid_sender, message_by_oid_receiver) = futures::channel::mpsc::channel(1);
```
Horusiath, line 200: Use `tokio::sync::mpsc::channel`: tokio is often more mature than futures. If generic constraints allow that you could also use `tokio::sync::oneshot`.

src/biz/collab/folder_view.rs
```diff
@@ -0,0 +1,162 @@
+use std::collections::HashSet;
+
+use collab_folder::Folder;
+use shared_entity::dto::workspace_dto::FolderView;
+use uuid::Uuid;
+
+pub fn collab_folder_to_folder_view(folder: &Folder, depth: u32) -> FolderView {
```
Horusiath, line 7: This looks like method that could be implemented directly on `FolderView`.

src/biz/workspace/publish.rs
```diff
@@ -0,0 +1,156 @@
...
+pub async fn publish_collabs(
+  pg_pool: &PgPool,
+  workspace_id: &Uuid,
+  publisher_uuid: &Uuid,
+  publish_items: Vec<PublishCollabItem<serde_json::Value, Vec<u8>>>,
+) -> Result<(), AppError> {
+  for publish_item in &publish_items {
+    check_collab_publish_name(publish_item.meta.publish_name.as_str())?;
+  }
+  insert_or_replace_publish_collabs(pg_pool, workspace_id, publisher_uuid, publish_items).await?;
```
Horusiath, line 27: Since there's no explicit Postgres transaction intiialization I can only assume that every sqlx query executed here will be wrapped and committed independently.

Another issue is that publish_items has no upper bound check. We basically copy unlimited number of collabs of unlimited size. With no timeouts.

src/biz/workspace/publish_dup.rs
```diff
@@ -0,0 +1,939 @@
...
+    }
+  }
+
+  async fn duplicate(mut self, publish_view_id: &str) -> Result<(), AppError> {
+    let mut txn = self.pg_pool.begin().await?;
+
+    // new view after deep copy
+    // this is the root of the document/database duplicated
+    let mut root_view = match self
+      .deep_copy(&mut txn, uuid::Uuid::new_v4().to_string(), publish_view_id)
```
Horusiath, line 109: It's hard to say how CPU intensive the deep copy is. Ideally we don't want to hold transactions for too long (as we have upper limit of concurrent connections in PgPool). 
1. If possible we could prepare blobs first, then begin transaction and insert them.
2. Alternatively if pt.1 uses too much memory, insert collabs one-by-one each one in separate transaction in their view parent order.

Maybe it would be a good idea to also add metrics to measure time spent on duplicating?

src/biz/workspace/publish_dup.rs
```diff
@@ -0,0 +1,939 @@
...
+          tracing::warn!("No published collab data found for view_id: {}", view_id);
+          return Ok(None);
+        },
+      };
+
+    let published_db = serde_json::from_slice::<PublishDatabaseData>(&published_blob)?;
+    let mut parent_view = self
+      .deep_copy_database_view(
+        txn,
+        uuid::Uuid::new_v4().to_string(),
```
appflowy, line 506: the same. can use gen_view_id

src/biz/workspace/publish_dup.rs
```diff
@@ -0,0 +1,939 @@
...
+    // collab of database
+    let mut db_collab = collab_from_doc_state(published_db.database_collab.clone(), "")?;
+    let pub_db_id = get_database_id_from_collab(&db_collab)?;
+
+    // check if the database is already duplicated
+    if let Some(db_id) = self.duplicated_refs.get(&pub_db_id).cloned().flatten() {
+      return Ok((pub_db_id, db_id, true));
+    }
+
+    let new_db_id = uuid::Uuid::new_v4().to_string();
```
appflowy, line 560: can use gen_database_id

src/biz/workspace/publish_dup.rs
```diff
@@ -0,0 +1,939 @@
...
+    self
+      .duplicated_refs
+      .insert(pub_db_id.clone(), Some(new_db_id.clone()));
+
+    // duplicate db collab rows
+    for (old_id, row_bin_data) in &published_db.database_row_collabs {
+      // assign a new id for the row
+      let new_row_id = uuid::Uuid::new_v4().to_string();
```
appflowy, line 568: gen_row_id

src/biz/workspace/publish_dup.rs
```diff
@@ -0,0 +1,939 @@
...
+      let mut db_views = view_map.get_all_views(&txn).into_iter().collect::<Vec<_>>();
+
+      for db_view in db_views.iter_mut() {
+        let new_view_id = if db_view.id == publish_view_id {
+          self
+            .duplicated_db_main_view
+            .insert(pub_db_id.clone(), new_view_id.clone());
+          new_view_id.clone()
+        } else {
+          uuid::Uuid::new_v4().to_string()
```
appflowy, line 634: gen_view_id

libs/database/src/publish.rs
speed2exe: @Horusiath iirc `{rows_affected}` account for both modified and inserted. my intent for every collab, it is either inserted or modified. I shall modify the code to log both  the `{rows_affected} and `{item_count}`, if they differ.

libs/database/src/publish.rs
speed2exe: there is an upper bound 4M for metadata, 128MB for blob. that was check before the request came in `post_publish_collabs_handler`. do you think these upper bounds are reasonable?

services/appflowy-collaborate/src/group/cmd.rs
speed2exe: i have tried. but based on the `subscribe` function signature, it's not possible to use the tokio channel

```rust
  pub async fn subscribe<Sink, Stream>(
    &self,
    user: &RealtimeUser,
    subscriber_origin: CollabOrigin,
    sink: Sink,
    stream: Stream,
  ) where
    Sink: SinkExt<CollabMessage> + Clone + Send + Sync + Unpin + 'static,
    Stream: StreamExt<Item = MessageByObjectId> + Send + Sync + Unpin + 'static,
    <Sink as futures_util::Sink<CollabMessage>>::Error: std::error::Error + Send + Sync,
```

src/biz/collab/folder_view.rs
speed2exe: apparently `FolderView` is in `shared-entity` crate to share structure between server and client. putting the implementation as where you mentioned will mean that the `shared-entity` crate will have to depend on the `collab-folder` for `Folder`, which we want to avoid having.

src/biz/workspace/publish.rs
speed2exe: there's only 1 sqlx operation for this api call, which is the `insert_or_replace_publish_collabs`. so there's no need for transaction. as mentioned above, upper bound is checked when api is call before it reaches the this part

services/appflowy-collaborate/src/group/cmd.rs
```diff
@@ -169,6 +182,40 @@ where
...
+        .subscribe(
+          user,
+          CollabOrigin::Server,
+          collab_message_sender,
+          message_by_oid_receiver,
+        )
+        .await;
+      let message = HashMap::from([(object_id, messages)]);
+      match message_by_oid_sender.try_send(message) {
+        Ok(()) => tracing::info!("sent message to group"),
```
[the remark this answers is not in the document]
speed2exe: changed to only logging with object_id when there is an error

services/appflowy-collaborate/src/group/cmd.rs
```diff
@@ -72,9 +75,19 @@ where
             collab_messages,
             ret,
           } => {
-            let result = self
-              .handle_client_collab_message(&user, object_id, collab_messages)
-              .await;
+            let result = match user.device_id.as_str() {
```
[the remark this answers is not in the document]
speed2exe: changed as suggested

services/appflowy-collaborate/src/command.rs
```diff
@@ -41,6 +51,27 @@ pub(crate) fn spawn_collaboration_command(
...
+                object_id,
+                collab_messages,
+                ret,
+              })
+              .await
+            {
+              tracing::error!("Send group command error: {}", err);
+            };
+          },
+          None => tracing::info!("No group sender for object_id: {}", object_id),
```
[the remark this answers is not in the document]
speed2exe: removed the logging

services/appflowy-collaborate/src/collab/storage.rs
```diff
@@ -404,4 +407,50 @@ where
...
+      .await
+      .map_err(|err| {
+        AppError::Unhandled(format!(
+          "Failed to send encode collab command to realtime server: {}",
+          err
+        ))
+      })?;
+
+    // Spawn a task to wait for the response from the realtime server
+    tokio::spawn(async move {
```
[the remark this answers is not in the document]
speed2exe: changed to wait for response

src/biz/workspace/publish_dup.rs
```diff
@@ -0,0 +1,939 @@
...
+      pg_pool,
+      collab_storage,
+      duplicator_uid: dest_uid,
+      dest_workspace_id,
+      dest_view_id,
+    }
+  }
+
+  async fn duplicate(mut self, publish_view_id: &str) -> Result<(), AppError> {
+    let mut txn = self.pg_pool.begin().await?;
```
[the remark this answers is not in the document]
speed2exe: I have remove the need for postgres txn for the most part of the duplicate operation, it now accumulate all the respective collabs before inserting into using  `collab_storage`

src/biz/workspace/publish_dup.rs
speed2exe: I have changed to accumulating all the collabs to insert without holding on to a postgres txn as you mentioned. We already have the metrics to measure the frequency and latency of each http endpoint (including this one), that would suffice for now. when we need more thorough analysis I'll add them to be included to the `/metrics`

libs/database/src/publish.rs
Horusiath: Tbh. I think that 128MB could be used to DDoS us. But we can adjust it accordingly in the future.

speed2exe: There's also a max limit of 256MB for a single request in our load balancer on this particular endpoint. in additional, this also consumes workspace storage space, so it would be a matter of tries before the user's limit is reached and will be blocked.

src/biz/workspace/publish_dup.rs
```diff
@@ -0,0 +1,959 @@
...
+          }
+          txn_wrapper.encode_update_v1()
+        };
+        let updated_ws_w_db_collab = collab_to_bin(&ws_db_collab, CollabType::WorkspaceDatabase);
+        (ws_db_updates, updated_ws_w_db_collab)
+      })
+      .await?;
+
+      collab_storage
+        .insert_new_collab_with_transaction(
```
appflowy, line 193: how about using `insert_new_collab`? it will save the collab in memory and the write to disk in the background.

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #702, opened 2024-07-25
feat: improve workspace api

speed2exe:
- deprecate the user of `AFWorkspace`, just use `Vec<AFWorkspace>`
- add option to include `member_count` for each workspace when fetching workspace

### DIFF
--- a/libs/client-api/src/http.rs
+++ b/libs/client-api/src/http.rs
@@ -1,5 +1,6 @@
 use crate::notify::{ClientToken, TokenStateReceiver};
 use app_error::AppError;
+use client_api_entity::workspace_dto::QueryWorkspaceParam;
 use client_api_entity::AuthProvider;
 use client_api_entity::CollabType;
 use gotrue::grant::PasswordGrant;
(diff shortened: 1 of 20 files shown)


libs/client-api/src/http.rs
```diff
@@ -627,16 +628,26 @@ impl Client {
     AppResponse::<()>::from_response(resp).await?.into_error()
   }
 
+  pub async fn get_workspaces(&self) -> Result<Vec<AFWorkspace>, AppResponseError> {
+    self
+      .get_workspaces_opt(QueryWorkspaceParam::default())
+      .await
+  }
+
   #[instrument(level = "info", skip_all, err)]
```
speed2exe, line 637: @qinluhe use for additional member count field, pass in `QueryWorkspaceParam` with `include_member_count` set to true

qinluhe:
LGTM

qinluhe (approved):
LGTM

libs/client-api/src/http.rs
```diff
@@ -627,16 +628,26 @@ impl Client {
...
-  pub async fn get_workspaces(&self) -> Result<AFWorkspaces, AppResponseError> {
```
appflowy, line 631: Is it compatible with old api endpoint?

speed2exe: yes it is backward compatible. frontend appflowy have to change from using `AFWorkspaces` to `Vec<Workspace>` when update, but that's just rust structure, the http endpoint behaviour is same for old requests

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #703, opened 2024-07-25
feat: endpoints for reactions on published view

khorshuheng:
Endpoints to support creating and removing reactions on published view.

### DIFF
--- a/libs/client-api/src/http_publish.rs
+++ b/libs/client-api/src/http_publish.rs
@@ -1,7 +1,7 @@
 use bytes::Bytes;
 use client_api_entity::{
-  CreateGlobalCommentParams, DeleteGlobalCommentParams, GlobalComments, PublishInfo,
-  UpdatePublishNamespace,
+  CreateGlobalCommentParams, CreateReactionParams, DeleteGlobalCommentParams, DeleteReactionParams,
+  GlobalComments, PublishInfo, Reactions, UpdatePublishNamespace,
 };
 use reqwest::Method;
 use shared_entity::response::{AppResponse, AppResponseError};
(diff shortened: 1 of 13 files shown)


qinluhe (approved):

libs/client-api/src/http_publish.rs
```diff
@@ -205,4 +249,24 @@ impl Client {
...
+    &self,
+    view_id: &uuid::Uuid,
+    comment_id: &Option<uuid::Uuid>,
+  ) -> Result<Reactions, AppResponseError> {
+    let url = format!(
+      "{}/api/workspace/published-info/{}/reaction",
+      self.base_url, view_id
+    );
+    let url = if let Some(comment_id) = comment_id {
+      format!("{}?comment_id={}", url, comment_id)
```
speed2exe, line 263: I'd prefer to use the `query` method before the `send` on the request http builder to pass in `comment_id` there instead of appending to url

libs/database/src/workspace.rs
```diff
@@ -1227,3 +1227,164 @@ pub async fn update_comment_deletion_status<'a, E: Executor<'a, Database = Postg
 
   Ok(())
 }
+
+#[derive(PartialEq, Eq, Hash)]
+struct ReactionKey {
+  comment_id: Uuid,
+  reaction_type: String,
```
speed2exe, line 1234: I think we should use enum for `reaction_type` instead?

libs/database/src/workspace.rs
```diff
@@ -1227,3 +1227,164 @@ pub async fn update_comment_deletion_status<'a, E: Executor<'a, Database = Postg
...
+    .collect();
+
+  Ok(reactions)
+}
+
+pub async fn select_reactions_for_comment<'a, E: Executor<'a, Database = Postgres>>(
+  executor: E,
+  comment_id: &Uuid,
+) -> Result<Vec<Reaction>, AppError> {
+  let rows = sqlx::query!(
```
speed2exe, line 1297: have you considered using `query_as!`? it might simplify what you're trying to do here

src/api/workspace.rs
```diff
@@ -1128,6 +1135,56 @@ async fn delete_published_collab_comment_handler(
   Ok(Json(AppResponse::Ok()))
 }
 
+#[derive(Deserialize)]
+struct GetReactionQuery {
+  comment_id: Option<Uuid>,
+}
+
```
speed2exe, line 1142: `GetReactionQuery` can be shared as a common param structure for `client-api`. it can be passed in from the client side in the reqwest `query` method.

libs/database/src/workspace.rs
khorshuheng: It's going to be difficult for us to define a list of enum for all possible emoji. Perhaps we can use the emoji crate?

speed2exe: hmm then I think it's fine, we can just stick to string

qinluhe: This isn't "type", it's emoji. It's just that when designing this interface, "type" is usually used as the attribute name because users of the same reaction type need to be aggregated.

libs/database/src/workspace.rs
khorshuheng: query_as! would not be suitable in this case, because multiple rows in the query result are fold to a single Reaction, as oppose to one row to one Reaction mapping.

khorshuheng: @speed2exe Just realized that i could have simplified the query by using agg_array function. Though `query_as` is still a little tricky to implement, because i will need to add some sqlx macro on the entity (eg. AFWebUser), but the entity library currently doesn't have sqlx as a dependency, and i am not sure if i should pull in sqlx dependency for the crate just for this.

Will need to think more about this, `query_as` is certainly a cleaner solution if we can use it.

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #721, opened 2024-08-12
feat: api endpoint for template categories and creators

khorshuheng:
CRUD API endpoints for template category and template creator.


libs/client-api/src/http_template.rs
```diff
@@ -0,0 +1,232 @@
...
+fn template_creator_resource_url(base_url: &str, creator_id: Uuid) -> String {
+  format!(
+    "{}/{}",
+    template_creator_resources_url(base_url),
+    creator_id
+  )
+}
+
+impl Client {
+  pub async fn create_template_category(
```
speed2exe, line 38: since there's many param, maybe can just have client to pass in `CreateTemplateCategoryParams` instead. this can also potentially avoid unneeded `to_string()`.

libs/database/src/pg_row.rs
```diff
@@ -275,3 +275,93 @@ impl From<AFReactionRow> for Reaction {
...
+
+impl From<AFTemplateCategoryRow> for TemplateCategory {
+  fn from(value: AFTemplateCategoryRow) -> Self {
+    Self {
+      id: value.id,
+      name: value.name,
+      icon: value.icon,
+      bg_color: value.bg_color,
+      description: value.description,
+      category_type: value.category_type.into(),
```
speed2exe, line 298: is it not possible to represent `category_type` from returned row in sqlx? i am thinking if the type conversion boilerplates can be avoided

libs/database/src/template.rs
```diff
@@ -0,0 +1,373 @@
...
+  substr_match: &str,
+) -> Result<Vec<TemplateCreator>, AppError> {
+  let creator_rows = sqlx::query_as!(
+    AFTemplateCreatorRow,
+    r#"
+    SELECT
+    tc.creator_id AS "id!",
+    name AS "name!",
+    avatar_url AS "avatar_url!",
+    ARRAY_AGG((al.link_type, al.url)) FILTER (WHERE link_type IS NOT NULL) AS "account_links: Vec<AccountLinkColumn>"
```
speed2exe, line 310: can indent the returned columns for better readability

migrations/20240806054557_template_category.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS af_template_category (
+  category_id   UUID NOT NULL DEFAULT gen_random_uuid(),
+  created_at    TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at    TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
speed2exe, line 4: maybe your intention is to use a trigger to update the `updated_at` field when there's change in the row. i dont see this field in the sqlx sql. if you dont need to the updated time, can also remove this column altogether

migrations/20240806103039_template_creator.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS af_template_creator (
+  creator_id        UUID NOT NULL DEFAULT gen_random_uuid(),
+  created_at        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
speed2exe, line 4: same as previous comment

migrations/20240806054557_template_category.sql
khorshuheng: It was my mistake, i didn't include it on the sqlx sql. As of the moment this field is not being used for frontend, though i do think there is a good possibility that that may change in the future.

libs/database/src/pg_row.rs
khorshuheng: Currently, the category type is being stored as int type in postgres. So one way to avoid such conversion would be to use integer type for category_type field as opposed to enum - though I would prefer not to in this case, since we only have a limited number of category type.

As far as i know it is not possible to automatically convert an integer to enum type without having a struct that derives sqlx type.

PR #736, opened 2024-08-22
feat: endpoint for avatar upload

khorshuheng:
Changes:
1. Add end point for supporting template center avatar upload
2. The bucket client was assuming that the object keys are always inferred from workspace id. As a result, it's hard to use the client for objects which is not associated with a workspace, such as avatar. This has been modified such that the callers of the trait are responsible for constructing the correct object key.
3. Instead of storing the content type via a separate metadata object, the content type was specified while uploading the blob to s3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -132,6 +132,44 @@ dependencies = [
 ]
 
 [[package]]
+name = "actix-multipart"
+version = "0.7.2"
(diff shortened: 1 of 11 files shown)


appflowy (approved):

appflowy:
LGTM

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #747, opened 2024-08-26
chore: drop group with maximum timeout

appflowy:
drop/disconnect the collaboration group if there are no active editing sessions and the group's lifetime has exceeded the maximum allowed duration.

### DIFF
--- a/services/appflowy-collaborate/src/group/cmd.rs
+++ b/services/appflowy-collaborate/src/group/cmd.rs
@@ -8,7 +8,7 @@ use tracing::{instrument, trace, warn};
 
 use access_control::collab::RealtimeAccessControl;
 use collab_rt_entity::user::RealtimeUser;
-use collab_rt_entity::{ClientCollabMessage, ServerCollabMessage, SinkMessage};
+use collab_rt_entity::{AckCode, ClientCollabMessage, ServerCollabMessage, SinkMessage};
 use collab_rt_entity::{CollabAck, RealtimeMessage};
 use database::collab::CollabStorage;
 
@@ -155,7 +155,11 @@ where
         let origin = first_message.origin().clone();
         let msg_id = first_message.msg_id();
         let object_id = first_message.object_id().to_string();
-        let ack = CollabAck::new(origin, object_id, msg_id, 0);
+
+        // when the group with given id is not found and the the first message is not init sync.
+        // Return AckCode::CannotApplyUpdate to the client and then client will send the init sync message.
+        let ack =
+          CollabAck::new(origin, object_id, msg_id, 0).with_code(AckCode::CannotApplyUpdate);
         entry
           .value()
           .send_message(ServerCollabMessage::ClientAck(ack).into())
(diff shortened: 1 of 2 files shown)


services/appflowy-collaborate/src/group/cmd.rs
```diff
@@ -155,7 +155,11 @@ where
         let origin = first_message.origin().clone();
         let msg_id = first_message.msg_id();
         let object_id = first_message.object_id().to_string();
-        let ack = CollabAck::new(origin, object_id, msg_id, 0);
+
+        // when the group with given id is not found and the the first message is not init sync.
+        // Return AckCode::CannotApplyUpdate to the client and then client will send the init sync message.
+        let ack =
+          CollabAck::new(origin, object_id, msg_id, 0).with_code(AckCode::CannotApplyUpdate);
```
Horusiath, line 162: Cannot we just close connection here and force client to reconnect?:

1. When connecting, client will send init sync message anyway.
2. There may be some expectations about what state the client is in when doing init sync with regards to things like sequence numbers etc. Resetting the connection makes sure that we start at init state.

services/appflowy-collaborate/src/group/group_init.rs
```diff
@@ -200,7 +200,26 @@ impl CollabGroup {
       modified_at.elapsed().as_secs() > 60 * 60
     } else {
       let elapsed_secs = modified_at.elapsed().as_secs();
-      elapsed_secs > self.timeout_secs() && self.subscribers.is_empty()
+      if elapsed_secs > self.timeout_secs() {
+        // Mark the group as inactive if it has been inactive for more than 3 hours, regardless of the number of subscribers.
+        // Otherwise, return `true` only if there are no subscribers remaining in the group.
+        // If a client modifies a group that has already been marked as inactive (removed),
+        // the client will automatically send an initialization sync to reinitialize the group.
+        const MAXIMUM_SECS: u64 = 3 * 60 * 60;
```
Horusiath, line 208: You don't need to use `u64`: Rust has `Duration` type specifically for type calculation, and `Duration::from_secs` is const fn so can be used with const values.

services/appflowy-collaborate/src/group/cmd.rs
appflowy: close websocket connect will force all collab objects to start init sync. But here I just force the corresponding collab to start init sync. For old version appflowy client, we can stick to this approach. For new version, we introduce new sync protocol

services/appflowy-collaborate/src/group/group_init.rs
appflowy: using `MAXIMUM_SECS` to compare with  `elapsed_secs`. So create an const instance here instead of using 3 * 60 * 60.

Horusiath: `modified_at.elapsed()` already returns `Duration` and durations can be compared against each other as well.

Horusiath (approved):

services/appflowy-collaborate/src/group/cmd.rs
Horusiath: What's the original reason behind the update that cannot be applied?

lilpoozie2005 (approved):

PR #750, opened 2024-08-26
add minimum client version validation in websocket connection establishment

Horusiath:
This PR adds minimal client version verification when establishing WS connection. It's configurable via `APPFLOWY_WEBSOCKET_CLIENT_MIN_VERSION` env var (defaults to `0.5.0`).


khorshuheng:
Client version older than 0.5.0 is still using the other endpoint: https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/src/api/ws.rs#L40

lilpoozie2005 (approved):

PR #769, opened 2024-08-30
chore: decode update when spawn blocking when update size exceeds the…

appflowy:
decode update when spawn blocking when update size exceeds the threshold size


appflowy:
> I'd say that 0.5MB may be too small. Assuming yrs benchmarks, we parse ~140KB in 2.3ms. I'd say that spawning a Rust thread is justifiable for work that takes at least 16ms (and that's already pretty small). So IMO threshold should be around 1-2MB. Only a handful of pathological document cases require that.

Ok. I will use 1MB.

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #773, opened 2024-08-30
chore: reconnect realtime metrics

Horusiath:
Reconnect realtime collab metrics, that were detached from the registry in #770

### DIFF
--- a/services/appflowy-collaborate/src/metrics.rs
+++ b/services/appflowy-collaborate/src/metrics.rs
@@ -1,5 +1,11 @@
+use std::sync::Arc;
+use std::time::Duration;
+
 use prometheus_client::metrics::gauge::Gauge;
 use prometheus_client::registry::Registry;
+use tokio::time::interval;
+
+use database::collab::CollabStorage;
 
 #[derive(Clone, Default)]
 pub struct CollabRealtimeMetrics {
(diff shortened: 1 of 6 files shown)


services/appflowy-collaborate/src/rt_server.rs
```diff
@@ -68,15 +68,14 @@ where
       info!("CollaborationServer with actix-web runtime");
     }
 
-    let metrics_calculate = CollabRealtimeMetrics::default();
```
Horusiath, line 71: This was the issue. I didn't notice that we already had an `Arc` reference to properly initialised `CollabRealtimeMetrics` and instead was creating new one, not connected to registry.

services/appflowy-collaborate/src/metrics.rs
```diff
@@ -85,6 +91,27 @@ impl CollabRealtimeMetrics {
   }
 }
 
+pub(crate) fn spawn_metrics<S>(metrics: Arc<CollabRealtimeMetrics>, storage: Arc<S>)
```
Horusiath, line 94: This was also removed in #770 but that part was not just copied from removed structure, it was periodically sampled and therefore should stay.

khorshuheng (approved):
LGTM

lilpoozie2005 (approved):

lilpoozie2005 (approved):

PR #782, opened 2024-09-02
chore: use colla storage instead of select from db directly

appflowy:
Use CollabStorage to get encode collab instead of query from db directly.


khorshuheng (approved):
lgtm

PR #792, opened 2024-09-04
feat: Support Railway Deployments

zahin-mohammad:
This PR makes some minor changes to support deploying on railway: 
- allows setting the host and port for the admin service via an environment variable. This allows using ipv6 (this was already done for af-cloud)
- exposes container ports for af-cloud and admin via environment variables

Testing: 
- Tested on my own deployments on railway

### DIFF
--- a/admin_frontend/src/config.rs
+++ b/admin_frontend/src/config.rs
@@ -2,21 +2,26 @@ use tracing::warn;
 
 #[derive(Debug, Clone)]
 pub struct Config {
+  pub host: String,
+  pub port: u16,
   pub redis_url: String,
   pub gotrue_url: String,
   pub appflowy_cloud_url: String,
 }
 
 impl Config {
-  pub fn from_env() -> Self {
-    Config {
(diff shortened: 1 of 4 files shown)


admin_frontend/src/config.rs
```diff
@@ -2,21 +2,26 @@ use tracing::warn;
...
   pub appflowy_cloud_url: String,
 }
 
 impl Config {
-  pub fn from_env() -> Self {
-    Config {
+  pub fn from_env() ->  Result<Config, anyhow::Error> {
+    let cfg = Config {
+      host: get_or_default("APPFLOWY_ADMIN_HOST", "0.0.0.0"),
+      port: get_or_default("APPFLOWY_ADMIN_PORT", "3000").parse()?,
```
speed2exe, line 16: for consistency, can we use `ADMIN_FRONTEND` as the prefix instead of `APPFLOWY_ADMIN`?

zahin-mohammad: Good idea.

PR #795, opened 2024-09-05
Feat/duplicate relation

speed2exe:


libs/client-api-test/src/test_client.rs
```diff
@@ -215,6 +216,37 @@ impl TestClient {
...
+          collab_type: CollabType::Database,
+        },
+      })
+      .await
+      .unwrap();
+    let db_doc_state = db_collab_collab_resp.encode_collab.doc_state;
+    Collab::new_with_source(
+      CollabOrigin::Server,
+      &db_id,
+      DataSource::DocStateV1(db_doc_state.to_vec()),
```
appflowy, line 243: can use `db_collab_collab_resp.encode_collab.into()` directly

libs/client-api-test/src/test_client.rs
```diff
@@ -215,6 +216,37 @@ impl TestClient {
...
 
+  pub async fn get_db_collab_from_view(&mut self, workspace_id: &str, view_id: &str) -> Collab {
+    let mut ws_db_collab = self.get_workspace_database_collab(workspace_id).await;
+    let ws_db_body = WorkspaceDatabaseBody::new(&mut ws_db_collab);
+    let txn = ws_db_collab.transact();
+    let db_id = ws_db_body
+      .get_all_database_meta(&txn)
+      .into_iter()
+      .find(|db_meta| db_meta.linked_views.contains(&view_id.to_string()))
+      .unwrap()
```
appflowy, line 227: remove this unwrap

speed2exe: this is in the test client. it should be fine?

PR #798, opened 2024-09-06
feat: Support s3 as published collab storage

khorshuheng:
This PR will introduce a new trait, PublishedCollabStore, which will handle the different operations for publish collabs. Two different implementation has been introduced: one write and read exclusively from postgres as per status quo, the other will write to both s3 and postgres, and read from s3 with postgres as fallback.

### DIFF
--- a/src/biz/workspace/publish.rs
+++ b/src/biz/workspace/publish.rs
@@ -1,67 +1,29 @@
+use std::sync::Arc;
+
 use app_error::AppError;
+use async_trait::async_trait;
 use database_entity::dto::{PublishCollabItem, PublishInfo};
+use shared_entity::dto::publish_dto::PublishViewMetaData;
 use sqlx::PgPool;
+use tracing::debug;
 use uuid::Uuid;
 
 use database::{
+  file::{s3_client_impl::AwsS3BucketClientImpl, BucketClient, ResponseBlob},
   publish::{
     delete_published_collabs, insert_or_replace_publish_collabs, select_publish_collab_meta,
     select_published_collab_blob, select_published_collab_info,
-    select_user_is_collab_publisher_for_all_views, select_workspace_publish_namespace,
-    select_workspace_publish_namespace_exists, update_workspace_publish_namespace,
+    select_published_collab_workspace_view_id, select_published_data_for_view_id,
+    select_published_metadata_for_view_id, select_user_is_collab_publisher_for_all_views,
+    select_workspace_publish_namespace, select_workspace_publish_namespace_exists,
+    update_workspace_publish_namespace,
   },
   workspace::select_user_is_workspace_owner,
 };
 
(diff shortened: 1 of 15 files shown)


src/biz/workspace/publish.rs
```diff
@@ -154,3 +120,296 @@ async fn check_workspace_namespace(new_namespace: &str) -> Result<(), AppError>
...
+    for publish_item in &publish_items {
+      check_collab_publish_name(publish_item.meta.publish_name.as_str())?;
+      let object_key = get_collab_s3_key(workspace_id, &publish_item.meta.view_id);
+      let data = publish_item.data.clone();
+      let bucket_client = self.bucket_client.clone();
+      let metrics = self.metrics.clone();
+      tokio::spawn(async move {
+        let result = bucket_client.put_blob(&object_key, &data).await;
+        if let Err(err) = result {
+          debug!("Failed to publish collab to S3: {}", err);
```
speed2exe, line 295: maybe we can create a `mut vec![]` to store all the handles returned by `tokio::spawn`, then await them all. this will ensure that if the endpoint returns success, we have indeed put all the data in S3 and avoid situation where a success code is returned to client, but there's error putting in S3.

speed2exe:
@khorshuheng if we switch to use s3 as backend storage for published data, do we have to import the existing published data in postgres into S3, or is there a fallback if we fail to fetch from S3?

src/biz/workspace/publish.rs
khorshuheng: Yeah, i was wondering whether we should do so. The downside of waiting for the request is, the time taken to make sure that the collab is written to both postgres and S3 will increase. Will need further discussion to determine whether the increase in latency is worth it or not, since we do have a fallback to postgres.

src/biz/workspace/publish.rs
```diff
@@ -154,3 +120,296 @@ async fn check_workspace_namespace(new_namespace: &str) -> Result<(), AppError>
...
+      Some((workspace_id, js_val)) => {
+        let metadata = serde_json::from_value(js_val)?;
+        let object_key = get_collab_s3_key(&workspace_id, view_id);
+        match self.bucket_client.get_blob(&object_key).await {
+          Ok(resp) => {
+            self.metrics.incr_success_read_count(1);
+            Ok(Some((metadata, resp.to_blob())))
+          },
+          Err(_) => {
+            let result = match select_published_data_for_view_id(&self.pg_pool, view_id).await? {
```
Horusiath, line 341: When sending collab state to S3, are we removing it from Postgres? Otherwise, I think we should check if we have Error::NotFound here. Without it, over transition period from Postgres&rarr;S3, we might get stale state from postgres.

khorshuheng: For now, the collab state are written to both postgres and s3 on every publish, so both s3 and postgres will have the same data.

PR #810, opened 2024-09-11
fix: batch create collab api endpoint might cause tokio runtime issue

appflowy:
The batch/collab API endpoint is taking too long to complete, primarily because the decompression step is causing performance issues with the Tokio runtime. To address this:

1. remove the old version of the API endpoint.
2. optimize the decompression process by offloading it to Rayon, which will handle the decompression in parallel, improving performance and preventing the Tokio runtime from being overwhelmed by CPU-bound tasks.


src/api/workspace.rs
```diff
@@ -583,68 +578,72 @@ async fn batch_create_collab_handler(
...
     }
   }
-  let duration = start_time.elapsed();
-  event!(
-    tracing::Level::DEBUG,
-    "end decompressing collab params list, time taken: {:?}",
-    duration
-  );
+  // Perform decompression and processing in a Rayon thread pool
+  let mut collab_params_list = tokio::task::spawn_blocking(move || {
```
Horusiath, line 613: You're still spawning a thread in here.

Horusiath (commented):
This is something that probably should be measured, at least in performance tests. Do we know how much size does Brotli saves us (after all it's optimised for HTML data not regular Yrs format), how long does it actually take and how other algorithms (ie. zstd) perform in comparison?

src/api/workspace.rs
appflowy: without using the spawn_blocking, it might block the tokio worker thread, causing the runtime to stall.

appflowy:
> This is something that probably should be measured, at least in performance tests. Do we know how much size does Brotli saves us (after all it's optimised for HTML data not regular Yrs format), how long does it actually take and how other algorithms (ie. zstd) perform in comparison?

Yes. we can. But old fontend clients are still using Brotli to compress data.  when import a csv file that contains lots of rows.Each of them is a collab. Maybe I can add some metric that determine using rayon or not.

PR #811, opened 2024-09-11
feat: add additional fields to FolderView and support different root id

khorshuheng:
Fixes:
- Originally, the default folder view is return if a view cannot be retrieved by id from the Folder struct. The appflowy web frontend might display the wrong result. This PR filter out such views from the output tree instead.
- There are child views which their parent view id attribute does not match the parent view's id. Such case is due to malformed encoded collab and should have been filtered out.

New features:
- A new query parameter named root_view_id, that allows the caller to iterate the folder view from any view, not just the root workspace id. Will be used in the future by the appflowy web frontend.
- Add is_published and layout fields to the FolderView.
- Return error if no valid nodes are found in the folder view tree

### DIFF
--- a/libs/app-error/src/lib.rs
+++ b/libs/app-error/src/lib.rs
@@ -139,6 +139,9 @@ pub enum AppError {
 
   #[error("{0}")]
   InvalidPublishedOutline(String),
+
+  #[error("{0}")]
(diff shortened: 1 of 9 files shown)


speed2exe:
LGTM

PR #816, opened 2024-09-12
Feat/dup refactor

speed2exe:
- Add test cases for duplicating inner doc in database rows
- Refactor to use api from Appflowy-Collab as much as possible

This PR requires https://github.com/AppFlowy-IO/AppFlowy-Collab/pull/283 and will be marked ready when it merges


libs/client-api-test/src/test_client.rs
```diff
@@ -193,23 +193,18 @@ impl TestClient {
...
+        workspace_id.to_string(),
+        db_storage_id.to_string(),
+        CollabType::WorkspaceDatabase,
+      )
       .await
-      .unwrap()
-      .encode_collab
-      .doc_state
-      .to_vec();
+      .unwrap();
```
appflowy, line 203: why unwrap here? Can we return a Result

src/biz/workspace/publish_dup.rs
```diff
@@ -740,52 +744,82 @@ impl PublishCollabDuplicator {
...
+
+        // duplicate row document if exists
+        if let Some(pub_row_doc_id) = pub_row_doc_id {
+          if let Some(row_doc_doc_state) = published_db
+            .database_row_document_collabs
+            .get(&pub_row_doc_id)
+          {
+            let pub_doc_collab =
+              collab_from_doc_state(row_doc_doc_state.to_vec(), &pub_row_doc_id)?;
+            let pub_doc =
```
appflowy, line 775: if fail to open the doucment, can we just ignore its error and continuous the following tasks

src/biz/workspace/publish_dup.rs
```diff
@@ -740,52 +744,82 @@ impl PublishCollabDuplicator {
...
 
+          // collect all cell with field type as relation
           let mut rel_row_idss = vec![];
           for (_, out) in cells.iter(&txn) {
             if let Ok(m) = out.cast::<MapRef>() {
-              if let Some(Out::Any(Any::BigInt(n))) = m.get(&txn, "field_type") {
+              if let Some(Out::Any(Any::BigInt(n))) = m.get(&txn, CELL_FIELD_TYPE) {
                 if n == FieldType::Relation as i64 {
-                  let relation_data = m.get(&txn, "data").ok_or_else(|| {
+                  let relation_data = m.get(&txn, CELL_DATA).ok_or_else(|| {
```
appflowy, line 822: deep copy database will be aborted when there is error. Are these errors can be ignored?

src/biz/workspace/publish_dup.rs
```diff
@@ -740,52 +744,82 @@ impl PublishCollabDuplicator {
       // assign a new id for the row
       let dup_row_id = gen_view_id();
```
appflowy, line 745: use gen_row_id()

src/biz/workspace/publish_dup.rs
speed2exe: this shouldn't happen as this is a fundamental assumption of the data in the doc. nonetheless, I'll ignore it and log the error

src/biz/workspace/publish_dup.rs
speed2exe: ok

PR #824, opened 2024-09-15
fix: filter out templates which has not been published

khorshuheng:
Currently, list templates, get template home page, and get template includes templates which has been unpublished. Originally, the reason for not deleting the templates when the view has not been published, is so that, if someone needs to make modification to the underlying view, there is no need to recreate the template. If a visitor attempts to visit a view that has been unpublished, the visitor will receive an error.

We are now modifying this behavior such that templates which has been unpublished will be automatically hidden, to deal with situations where a view might be corrupted and not usable.


speed2exe (approved):
LGTM

PR #831, opened 2024-09-17
feat: add endpoint for getting page view collab

khorshuheng:
Add endpoints to support rendering a collab on the web version of Appflowy. The collab for the database rows are included together as part of the response, if the view layout is Grid, Board, or Calendar.

The current number of rows supported by the import CSV function is 1000 rows. Based on test, the performance for retrieving all 1000 rows at once is still acceptable, as long as there are not too many columns.

The yrs doc state is returned directly, as opposed to EncodedCollab.

### DIFF
--- a/src/biz/workspace/page_view.rs
+++ b/src/biz/workspace/page_view.rs
@@ -1,0 +1,242 @@
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+
+use app_error::ErrorCode;
+use appflowy_collaborate::collab::storage::CollabAccessControlStorage;
+use chrono::DateTime;
+use collab::core::collab::Collab;
+use collab_database::{
+  database::DatabaseBody, rows::RowId, workspace_database::WorkspaceDatabaseBody,
+};
+use collab_entity::{CollabType, EncodedCollab};
+use collab_folder::CollabOrigin;
+use database::collab::{select_workspace_database_oid, CollabStorage, GetCollabOrigin};
+use database::publish::select_published_view_ids_for_workspace;
+use database::user::select_web_user_from_uid;
+use database_entity::dto::{QueryCollab, QueryCollabResult};
+use rayon::iter::{IntoParallelIterator, ParallelIterator};
+use shared_entity::dto::workspace_dto::{FolderView, PageCollab, PageCollabData};
+use shared_entity::response::AppResponseError;
+use sqlx::PgPool;
+use uuid::Uuid;
+
+use crate::biz::collab::folder_view::{
+  parse_extra_field_as_json, to_dto_view_icon, to_view_layout,
+};
+use crate::biz::collab::{
+  folder_view::view_is_space,
+  ops::{get_latest_collab_encoded, get_latest_collab_folder},
+};
+
+use super::publish_dup::collab_from_doc_state;
+
+pub async fn get_page_view_collab(
+  pg_pool: &PgPool,
+  collab_access_control_storage: Arc<CollabAccessControlStorage>,
+  uid: i64,
+  workspace_id: Uuid,
+  view_id: &str,
+) -> Result<PageCollab, AppResponseError> {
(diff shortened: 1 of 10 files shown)


src/biz/workspace/page_view.rs
```diff
@@ -0,0 +1,229 @@
...
+    .iter()
+    .map(|row_id| QueryCollab {
+      object_id: row_id.to_string(),
+      collab_type: CollabType::DatabaseRow,
+    })
+    .collect();
+  let row_query_collab_results = collab_access_control_storage
+    .batch_get_collab(&uid, queries)
+    .await;
+  let row_collabs: HashMap<String, Vec<u8>> = row_query_collab_results
```
appflowy, line 186: it might have performance issue (cpu bound task) when there are lots of collab needed to deoced. Consider using rayon

src/biz/workspace/page_view.rs
```diff
@@ -0,0 +1,229 @@
...
+    .collect();
+  let queries: Vec<QueryCollab> = row_ids
+    .iter()
+    .map(|row_id| QueryCollab {
+      object_id: row_id.to_string(),
+      collab_type: CollabType::DatabaseRow,
+    })
+    .collect();
+  let row_query_collab_results = collab_access_control_storage
+    .batch_get_collab(&uid, queries)
```
appflowy, line 184: I am thinking about refactor the `batch_get_collab` that returns a collab stream instead.

src/biz/workspace/page_view.rs
```diff
@@ -0,0 +1,229 @@
...
+    GetCollabOrigin::User { uid },
+    &workspace_id.to_string(),
+    &db_oid,
+    CollabType::Database,
+  )
+  .await?;
+  let db_collab = Collab::new_with_source(
+    CollabOrigin::Server,
+    &db_oid,
+    DataSource::DocStateV1(db.doc_state.to_vec()),
```
appflowy, line 154: can use `db.into()`.

src/biz/workspace/page_view.rs
```diff
@@ -0,0 +1,229 @@
...
+    collab_access_control_storage.clone(),
+    GetCollabOrigin::User { uid },
+    &workspace_id.to_string(),
+    &ws_db_oid,
+    CollabType::WorkspaceDatabase,
+  )
+  .await?;
+  let mut ws_db_collab = collab_from_doc_state(ws_db.doc_state.to_vec(), &ws_db_oid)?;
+  let ws_db_body = WorkspaceDatabaseBody::open(&mut ws_db_collab);
+  let txn = ws_db_collab.transact();
```
appflowy, line 135: how about this? 
```rust
let db_oid = {
ws_db_collab.transact();
  let db_oid = ws_db_body
    .get_database_meta_with_view_id(&txn, view_id)
    .ok_or(AppResponseError::new(
      ErrorCode::NoRequiredData,
      format!("Database view {} not found", view_id),
    ))?
    .database_id
}
````

File: src/biz/workspace/page_view.rs [hunk 5]
```diff
@@ -0,0 +1,229 @@
...
+      ErrorCode::Internal,
+      format!(
+        "Unable to create collab from object id {}: {}",
+        &db_oid, err
+      ),
+    )
+  })?;
+  let db_body = DatabaseBody::from_collab(&db_collab).unwrap();
+  let txn = db_collab.transact();
+  let inline_view_id = db_body.get_inline_view_id(&txn);
```
appflowy, line 169: I prefer using following style. it makes the scope of txn more clear.
```rust
let inline_view_id = {
db_collab.transact();
db_body.get_inline_view_id(&txn)
};
```

src/biz/workspace/page_view.rs
khorshuheng: `db.into()` would cause `db` to be moved, which makes it not possible to call db.doc_state later.

PR #834, opened 2024-09-20
feat: support protobuf serialization and deserialization for CollabParams

khorshuheng:
Add methods to support conversion from CollabParams to protobuf serialized format and vice versa

### DIFF
--- a/libs/database-entity/src/dto.rs
+++ b/libs/database-entity/src/dto.rs
@@ -1,8 +1,12 @@
+use crate::error::EntityError;
+use crate::error::EntityError::{DeserializationError, InvalidData};
 use crate::util::{validate_not_empty_payload, validate_not_empty_str};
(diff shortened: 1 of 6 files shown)


libs/database-entity/src/dto.rs
```diff
@@ -107,6 +111,44 @@ impl CollabParams {
...
+        .as_ref()
+        .map(|embeddings| embeddings.to_proto()),
+    }
+  }
+
+  pub fn to_protobuf_bytes(&self) -> Vec<u8> {
+    self.to_proto().encode_to_vec()
+  }
+
+  pub fn from_proto(proto: proto::collab::CollabParams) -> Result<Self, EntityError> {
```
appflowy, line 131: Can we use `TryFrom` or `From`? then we can write 
```rust
 let collab_type = CollabType::from(proto.collab_type);
```

PR #835, opened 2024-09-20
Endpoint for server info

khorshuheng:
There are a few scenarios which it will be helpful for a client to be aware of the functionality currently supported by AppFlowy Cloud:

1. A user is using self hosting an older version of AppFlowy Cloud, which does not support some of the newer functionality introduced by the newer client. Having a server info endpoint will allow the client to either provide a fallback, or provide appropriate error message to the user.

2. A user is using an older version of AppFlowy client, which relies on functionality that is already deprecated by AppFlowy Cloud. By checking the features supported by the server, the client can prompt the user to upgrade the client version.

### DIFF
--- a/libs/client-api/src/http.rs
+++ b/libs/client-api/src/http.rs
@@ -1,6 +1,8 @@
 use crate::notify::{ClientToken, TokenStateReceiver};
 use app_error::AppError;
+use app_error::ErrorCode;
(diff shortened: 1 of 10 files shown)


khorshuheng:
The other alternative to returning a supported feature list, would be to return the server version, and the minimum supported client version. It's not possible to do this at the moment, however, because our server package version is always `0.1.0`.

PR #837, opened 2024-09-24
feat: support fetching encoded collab in batch from memory

khorshuheng:
Currently, there is no options for retrieving encoded collab in batch from the in-memory collab, unlike the fetch single collab counter part.

This is less of a problem for the flutter app, because we do have a websocket client which will eventually sync every collab to the latest, even if the initially fetched data might be outdated.

For AppFlowy Web, however, the changes are pull when the data is refreshed, and there is no websocket client there to keep the state of the collab to the latest. Hence, in cases where the user has a database, changes to the content of the database will not be reflected immediately on the web version of AppFlowy, because the content is fetched via the batch get encoded collab method.

Therefore, an option has been added to the batch get encoded collab method, for the storage class. The existing API will still behave as status quo, and doesn't fetch the encoded collab from in-memory. The get page view API will be affected.

### DIFF
--- a/services/appflowy-collaborate/src/collab/storage.rs
+++ b/services/appflowy-collaborate/src/collab/storage.rs
@@ -7,6 +7,7 @@ use collab::entity::EncodedCollab;
 use collab_entity::CollabType;
 use collab_rt_entity::ClientCollabMessage;
 use itertools::{Either, Itertools};
+use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use sqlx::Transaction;
 use tokio::time::timeout;
 use tracing::warn;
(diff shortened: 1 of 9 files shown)


services/appflowy-collaborate/src/collab/storage.rs
```diff
@@ -340,8 +375,41 @@ where
...
+            .map(|(object_id, encoded_collab)| {
+              let query_collab_result = QueryCollabResult::Success {
+                encode_collab_v1: encoded_collab.encode_to_bytes().unwrap(),
+              };
+              (object_id.clone(), query_collab_result)
+            })
+            .collect()
+        })
+        .await
+        .unwrap();
```
appflowy, line 401: should we return an error instead of unwrap?

services/appflowy-collaborate/src/command.rs
```diff
@@ -50,6 +68,35 @@ pub(crate) fn spawn_collaboration_command(
...
         },
+        CollaborationCommand::BatchGetEncodeCollab { object_ids, ret } => {
+          if let Some(group_manager) = weak_groups.upgrade() {
+            let tasks = futures::stream::iter(object_ids)
+              .map(|object_id| {
+                let cloned_group_manager = group_manager.clone();
+                tokio::task::spawn(async move {
+                  let group = cloned_group_manager.get_group(&object_id).await;
+                  if let Some(group) = group {
+                    (object_id, group.encode_collab().await.ok())
```
appflowy, line 79: it might be an issue when encoding a `big` collab. these futures might run concurrently, so i am thinking about using stream

services/appflowy-collaborate/src/collab/storage.rs
```diff
@@ -340,8 +375,41 @@ where
...
+      let editing_results = self
+        .batch_get_encode_collab_from_editing(
+          editing_queries
+            .iter()
+            .map(|q| q.object_id.clone())
+            .collect(),
+        )
+        .await;
+      let editing_query_collab_results: HashMap<String, QueryCollabResult> =
+        tokio::task::spawn_blocking(move || {
```
appflowy, line 389: maybe you can try to use rayon. we already use rayon for cpu bound tasks

PR #840, opened 2024-09-24
feat: implement import notion zip file api endpoint

appflowy:
This feature introduces a new API endpoint for importing a Notion-exported zip file. The import process involves scheduling background tasks via Redis, and a new service called `appflowy-worker` will handle the actual processing of these import tasks.

### Workflow:
1. **User uploads Notion zip file via the new API endpoint**.
2. **The API schedules an import task by adding it to a Redis stream group**.
3. **The `appflowy-worker` service**, running in the background, picks up the task and starts processing the import (e.g., unzipping, parsing, and inserting data into the system).
4. **Once the import is complete**, the status of the task is updated, and the user can check the task progress or completion status via other API endpoints.

Notity user via email will be implemented in another PR.


deploy.env
```diff
@@ -144,5 +144,9 @@ APPFLOWY_INDEXER_REDIS_URL=redis://redis:6379
 APPFLOWY_COLLABORATE_MULTI_THREAD=false
 APPFLOWY_COLLABORATE_REMOVE_BATCH_SIZE=100
 
+# AppFlowy Worker
+APPFLOWY_WORKER_REDIS_URL=redis://redis:6379
+APPFLOWY_WORKER_DATABASE_URL=postgres://postgres:password@localhost:5432/postgres
```
khorshuheng, line 149: I would assume that it should have been `postgres` as opposed to localhost.

docker-compose.yml
```diff
@@ -152,6 +152,24 @@ services:
       - APPFLOWY_HISTORY_ENVIRONMENT=production
       - APPFLOWY_HISTORY_DATABASE_URL=${APPFLOWY_HISTORY_DATABASE_URL}
 
+  appflowy_worker:
+    restart: on-failure
+    image: appflowyinc/appflowy_worker:${APPFLOWY_HISTORY_VERSION:-latest}
```
khorshuheng, line 157: A different variable can be used, as opposed to APPFLOWY_HISTORY_VERSION.

services/appflowy-worker/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+COPY . .
+
+WORKDIR /app/services/appflowy-worker
+RUN cargo build --release --bin appflowy_worker
+
+FROM debian:bookworm-slim AS runtime
+WORKDIR /app/services/appflowy-worker
+RUN apt-get update -y \
+    && apt-get install -y --no-install-recommends openssl \
+    && apt-get install protobuf-compiler -y \
```
khorshuheng, line 31: We don't have to install protobuf compiler in run time.

deploy.env
speed2exe: I'm thinking we should just reuse the env var from appflowy-cloud. I cant think of any situation where it should be different

libs/database/Cargo.toml
```diff
@@ -21,12 +21,12 @@ serde_json.workspace = true
...
-  "macros",
-  "runtime-tokio-rustls",
-  "rust_decimal",
+    "postgres",
+    "chrono",
+    "uuid",
+    "macros",
+    "runtime-tokio-rustls",
+    "rust_decimal",
 ] }
```
speed2exe, line 30: @appflowy @khorshuheng I remember seeing PR having indentation spaces change from 2 to 4, and now back to 2. we should keep it consistent.

appflowy: I've set the tab_spaces =2 In rustfmt.toml. But i am not sure the cargo.toml will use the rustfmt.toml configuration

khorshuheng: It was due to editor auto formatting. We can choose either 2 or 4, i can adjust my editor accordingly.

The main problem is that cargo fmt doesn't support formatting Cargo.toml, unfortunately, though there have been attempts to introduce this.

migrations/20240930135712_import_data.sql
```diff
@@ -0,0 +1,21 @@
+-- Add migration script here
+CREATE TABLE af_import_task(
+    task_id UUID NOT NULL PRIMARY KEY,
+    file_size BIGINT NOT NULL,          -- File size in bytes, BIGINT for large files
+    workspace_id TEXT NOT NULL,         -- Workspace id
+    created_by BIGINT NOT NULL,         -- User ID
+    status INT NOT NULL,                -- Status of the file import (e.g., 0 for pending, 1 for completed, 2 for failed)
```
speed2exe, line 7: SMALLINT would be preferred

appflowy: ok

services/appflowy-worker/Dockerfile
appflowy: `lib/database` crate depends on` lib/tonic-proto` create that needs `proto`

khorshuheng: Yes, we do require protobuf during build process. But in debian:bookworm-slim , where we are running the compiled binary, we don't require the compiler, it is only required for the build process.

libs/database/Cargo.toml
appflowy: it seems like my IDE auto format the cargo.toml file. I will try set tab_spaces = 2. Let's keep 2

appflowy: I found it.

image

xtask/src/main.rs
```diff
@@ -33,12 +35,24 @@ async fn main() -> Result<()> {
...
     .context("Failed to start AppFlowy-History process")?;
 
+  let mut appflowy_worker_cmd = Command::new("cargo")
+    .args([
+      "run",
+      "--manifest-path",
+      "./services/appflowy-worker/Cargo.toml",
+    ])
+    .spawn()
+    .context("Failed to start AppFlowy-History process")?;
```
khorshuheng, line 45: Should be AppFlowy-Worker instead.

services/appflowy-worker/src/application.rs
```diff
@@ -0,0 +1,174 @@
...
+    state.pg_pool.clone(),
+    state.redis_client.clone(),
+    Arc::new(state.s3_client.clone()),
+    Arc::new(email_notifier),
+    "import_task_stream",
+    30,
+  ));
+
+  let app = Router::new()
+    .route("/health", get(health_check))
```
speed2exe, line 106: I dont think health check is needed, since it's a background task consumer and should not any user http. Then we can remove the whole axum dependency/setup

appflowy: Do we need this endpoint to check the worker server is up or not?

khorshuheng: Consider that this worker is independent and no other services needs to know whether the worker server is up or not, i would say that the healthcheck endpoint has very limited use.

khorshuheng: If we really need to monitor whether the worker is performing well or not exposing prometheus metric is probably a better way then having a health check endpoint.

PR #843, opened 2024-09-26
Endpoints to support page view approval workflow

khorshuheng:
This PR provides the following endpoints:
- POST /api/access-request : for user to submit request to access a view
- GET /api/access-request/{request-id}: for user to retrieve information of an access request. Used for the approval page on AppFlowy Web.
- POST /api/access-request/{request-id}/approve: for workspace owner to add the requester as a member of a workspace

An email should be sent to the workspace owner when a user request for access. This is not implemented yet, will be in another PR.

### DIFF
--- a/migrations/20240924045045_access_request.sql
+++ b/migrations/20240924045045_access_request.sql
@@ -1,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS af_access_request (
+  request_id UUID NOT NULL DEFAULT gen_random_uuid(),
+  uid BIGINT NOT NULL REFERENCES af_user(uid) ON DELETE CASCADE,
+  workspace_id UUID NOT NULL,
+  view_id UUID NOT NULL,
+  status INT NOT NULL,
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  UNIQUE (uid, workspace_id, view_id),
+  PRIMARY KEY(request_id)
+);
+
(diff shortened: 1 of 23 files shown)


migrations/20240924045045_access_request.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS af_access_request (
+  request_id UUID NOT NULL DEFAULT gen_random_uuid(),
+  uid BIGINT NOT NULL REFERENCES af_user(uid) ON DELETE CASCADE,
+  workspace_id UUID NOT NULL,
+  view_id UUID NOT NULL,
+  approved BOOLEAN NOT NULL DEFAULT FALSE,
```
speed2exe, line 6: imo we should have at least 3 states for status, pending/approve/rejected (default pending). in that way when access is rejected, it is easier to query for the rest of pending access request

src/biz/access_request/ops.rs
```diff
@@ -0,0 +1,96 @@
...
+  folder_view::{to_dto_view_icon, to_view_layout},
+  ops::get_latest_collab_folder,
+};
+
+pub async fn create_access_request(
+  pg_pool: &PgPool,
+  workspace_id: Uuid,
+  view_id: Uuid,
+  uid: i64,
+) -> Result<Uuid, AppError> {
```
speed2exe, line 28: maybe we can a check here if access request already exists?

khorshuheng: Yes, we will return an error code specifically for existing access request if one is found. It is in the test case.

PR #848, opened 2024-09-28
feat: Send email to owner when user request workspace access

khorshuheng:
Add the functionality to send email to owner when user request workspace access.

This PR also:
1. Make the function of sending email templates more generic.
2. Add owner email as part of workspace row entity.


email_template/config.production.js
```diff
@@ -15,15 +15,17 @@ module.exports = {
...
     },
   },
   locals: {
-    cdnBaseUrl: 'https://raw.githubusercontent.com/AppFlowy-IO/AppFlowy-Cloud/main/assets/mailer_templates/build_production/',
+    cdnBaseUrl:
+      "https://raw.githubusercontent.com/AppFlowy-IO/AppFlowy-Cloud/main/assets/mailer_templates/build_production/",
     userIconUrl: "{{ user_icon_url }}",
     userName: "{{ username }}",
     acceptUrl: "{{ accept_url }}",
+    approveUrl: "{{ accept_url }}",
```
qinluhe, line 28: approveUrl: "{{ approve_url }}",

libs/database/src/workspace.rs
```diff
@@ -671,6 +672,7 @@ pub async fn select_workspace<'a, E: Executor<'a, Database = Postgres>>(
         database_storage_id,
         owner_uid,
         (SELECT name FROM public.af_user WHERE uid = owner_uid) AS owner_name,
+        (SELECT email FROM public.af_user WHERE uid = owner_uid) AS owner_email,
```
speed2exe, line 675: since we are using more than 2 fields in `af_user`, it might be better to use a JOIN

src/biz/access_request/ops.rs
```diff
@@ -16,18 +16,56 @@ use shared_entity::dto::access_request_dto::{AccessRequest, AccessRequestView};
...
+    appflowy_web_url, request_id
+  );
+  let email = access_request.workspace.owner_email.clone();
+  let recipient_name = access_request.workspace.owner_name.clone();
+  // use default icon until we have workspace icon
+  let workspace_icon_url =
+    "https://miro.medium.com/v2/resize:fit:2400/1*mTPfm7CwU31-tLhtLNkyJw.png".to_string();
+  let user_icon_url =
+    "https://cdn.pixabay.com/photo/2015/10/05/22/37/blank-profile-picture-973460_1280.png"
+      .to_string();
```
speed2exe, line 49: I think we can find the icons in the `metadata` column of `af_user` and `icon` column of `af_workspace`, fallback this if not found

khorshuheng: One problem with this approach, is that the url metadata currently being stored on the user profile is not actually an url, but emoji UTF-8 string. We will properly need proper image support before we can do this.

Issue #791, opened 2024-09-04
[FR] Ease up deployment


DenuxPlays:
I think a GitHub workflow that creates and publishes all the images needed will ease up deployment.
No need to build the database images and no need to clone the repo.
Just a simple compose file and the .env file.
This way also things like watchtower would work correctly.

**1~3 main use cases of the proposed feature**

Currently you need to clone the repo and build your own docker images in order to deploy.
That means an increased deployed and update complexity that is not quite necessary.

- easier deployment
- easier to update 
- better version management 
- no need for git

**what types of users can benefit from using your proposed feature**

Only selfhosters will benefit from it.

**Additional context**

I can make these workflows and create a pr if wanted.

khorshuheng:
For postgres image, we actually already have a workflow that push the image to dockerhub, under appflowyinc.

That being said, since we only use pgvector extension at the moment, we can also use pgvector/pgvector:pg16.

Similarly, we do build the gotrue image and pushed it, though we can also use `supabase/gotrue:v2.159.1`  .

Unfortunately, we still need to clone the repo. It is not because we need to build the image, we still need the following:
1. https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/migrations/before/20230312043000_supabase_auth.sql
2. https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/nginx/nginx.conf

Though the 2nd item could be optional with some modification on docker compose, and using nginx/traefik outside docker compose instead.

DenuxPlays:
Ah okay.
I just copied the files (for now) and I cannot get it to work.

I am currently stuck here: https://github.com/AppFlowy-IO/AppFlowy-Cloud/issues/823

Also I cannot reach minio with the default configuration.

almereyda:
I would like to assist with reviewing or preparing work on GitHub Actions for building the images.

Also, there were further gotrue migration regressions recently.

- #839

A Compose setup that appeared to be more stable had been documented to:

- #622 

Easing the deployment also means to support disabling unwanted or possibly unavailable ternary integrations. It would be sad to see OpenAI as a hard proprietary dependency.

- #565

almereyda:
> Also I cannot reach minio with the default configuration.

@DenuxPlays Would you like to open a separate issue to investigate this further?

khorshuheng:
[...]

PR #851, opened 2024-09-30
chore: use pgvector for postgres image instead of building

khorshuheng:
Partially address https://github.com/AppFlowy-IO/AppFlowy-Cloud/issues/791 . As we are currently only using pgvector extension and not expecting to use additional extensions in the near future, there's little reason not to use pgvector image instead.

The alternative to this would be to publish our custom postgres image to dockerhub instead, so that the self hosting users do not have to build the postgres image when they need to install appflowy cloud.


Horusiath (approved):
I'm all for that. If our requirements will change in the future, we may go back to our private Postgres Dockerfile approach.

PR #853, opened 2024-10-01
Stateless collab group

Horusiath:


Cargo.toml
```diff
@@ -62,31 +62,31 @@ bytes = "1.5.0"
...
 ] }
 tracing-bunyan-formatter = "0.3.9"
 sqlx = { workspace = true, default-features = false, features = [
-  "runtime-tokio-rustls",
-  "macros",
-  "postgres",
-  "uuid",
-  "chrono",
-  "migrate",
+    "runtime-tokio-rustls",
```
appflowy, line 84: In order to make it consistency. You can modify your IDE setting as shown below
image

libs/collab-stream/src/lease.rs
```diff
@@ -0,0 +1,143 @@
...
+      .invoke_async(&mut conn)
+      .await?;
+    Ok(result == 1)
+  }
+}
+
+impl Drop for LeaseAcquisition {
+  fn drop(&mut self) {
+    if let Some(conn) = self.conn.take() {
+      tokio::spawn(Self::release_internal(
```
appflowy, line 48: Do we need to care the `Result` from `release_internal`

libs/collab-stream/src/model.rs
```diff
@@ -355,8 +367,266 @@ impl TryFrom<CollabUpdateEvent> for StreamBinary {
...
+  #[inline]
+  fn from(value: u8) -> Self {
+    UpdateFlags(value)
+  }
+}
+
+impl Display for UpdateFlags {
+  fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+    if self.is_v2_encoded() {
+      write!(f, ".v1")?;
```
appflowy, line 589: would it be `v2`

libs/collab-stream/src/model.rs
```diff
@@ -355,8 +367,266 @@ impl TryFrom<CollabUpdateEvent> for StreamBinary {
...
+  }
+}
+
+#[repr(transparent)]
+#[derive(Copy, Clone, Eq, PartialEq, Default)]
+pub struct UpdateFlags(u8);
+
+impl UpdateFlags {
+  /// Flag bit to mark if update is encoded using [EncoderV2] (if set) or [EncoderV1] (if clear).
+  pub const IS_V2_ENCODED: u8 = 0b0000_0001;
```
appflowy, line 549: how about adding the v1_encode:

```rust
  pub const IS_V2_ENCODED: u8 = 0b0000_0001;
  pub const IS_V1_ENCODED: u8 = 0b0000_0100;
  pub const IS_COMPRESSED: u8 = 0b0000_0010;
```

script/run_local_server.sh
```diff
@@ -53,7 +53,7 @@ then
   cargo sqlx prepare --workspace
 fi
 
-cargo run --package xtask
+cargo run --release --package xtask
```
appflowy, line 56: do we need release build in local development?

tests/collab/stress_test.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use super::util::TestScenario;
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn run_multiple_text_edits() {
+  let test_scenario = Arc::new(TestScenario::open(
+    "./tests/collab/asset/automerge-paper.json.gz",
+  ));
+  let mut tasks = Vec::new();
+  for _i in 0..1 {
```
appflowy, line 19: do we need to increase the number?

libs/collab-stream/src/model.rs
Horusiath: These are bit flags. If `IS_V2_ENCODED` is not set, it implies V1 encoding.

services/appflowy-collaborate/src/group/state.rs
```diff
@@ -139,13 +143,6 @@ impl GroupManagementState {
       error!("Failed to insert close event to control stream: {}", err);
     }
 
-    if let Some(entry) = entry {
-      let group = entry.1;
-      group.stop().await;
```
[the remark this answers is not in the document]
Horusiath: We don't need it. Dropping `CollabGroup` triggers cancellation token which will break loops running in background tasks.

tests/collab/stress_test.rs
Horusiath: It's still not passing, I'm running into issues with ACKs. What's worse none of the present tests found these issues.

libs/collab-stream/src/client.rs
```diff
@@ -66,29 +84,143 @@ impl CollabRedisStream {
...
+      let until = if live {
+        None
+      } else {
+        let mut reply: StreamRangeReply = conn.xrevrange_count(&stream_key, "+", "-", 1).await?;
+        match reply.ids.pop().map(|stream_id| stream_id.id) {
+          None => Some(MessageId::default()),
+          Some(id) => Some(MessageId::try_from(id)?),
+        }
+      };
+      while until.is_none() || since < until.unwrap() {
```
appflowy, line 124: it looks more safer by using
```rust
while until.map_or(true, |u| since < u) {}
```

Horusiath: I made it other (arguably better) way that doesn't require pre-fetching final MessageId.

libs/collab-stream/src/model.rs
```diff
@@ -355,8 +367,264 @@ impl TryFrom<CollabUpdateEvent> for StreamBinary {
...
+
+  /// Returns Redis stream key, that's storing entries mapped to/from [CollabStreamUpdate].
+  pub fn stream_key(workspace_id: &str, object_id: &str) -> String {
+    // use `:` separator as it adheres to Redis naming conventions
+    format!("af:{}:{}:updates", workspace_id, object_id)
+  }
+
+  pub fn into_update(self) -> Result<collab::preclude::Update, StreamError> {
+    let bytes = if self.flags.is_compressed() {
+      zstd::decode_all(std::io::Cursor::new(self.data))?
```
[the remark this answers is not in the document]
Horusiath: Potentially, but currently this flag is not used yet.

Horusiath:
There's still an issue with awareness not being reached by the clients.

PR #859, opened 2024-10-04
feat: refactor access control to enable swapping implementation

khorshuheng:
- Consolidate all casbin specific implementation to an optional mod in access-control crate. This simplifies the process of replacing access control implementation with another (eg. external authorization service instead of casbin, or a simpler implementation that is purely based on postgres).
- Remove is_access_control_enabled from the access control crate, for better separation of concern. Whether access control should be enabled or not should be the responsibility of the caller, not the implementation of access control itself.
- Having a configuration that enable or disable access control within the implementation also complicates the implementation since we have to make sure that every single method required by the trait does not perform access control when access control is disabled.
- Use dynamic dispatch as opposed to generics, so that we can switch implementation of access control at run time (eg. use NoOps access control is access control is disabled).


libs/access-control/src/casbin/access.rs
```diff
@@ -80,30 +80,22 @@ impl AccessControl {
...
-      self.enforcer.remove_policy(uid, obj).await?;
-      let _ = self.change_tx.send(AccessControlChange::RemovePolicy {
-        uid: *uid,
-        oid: obj.object_id().to_string(),
-      });
-      Ok(())
-    } else {
-      Ok(())
-    }
+    self.enforcer.remove_policy(uid, obj).await?;
```
appflowy, line 93: would it be better if `update_policy` and `remove_policy` return an option value. Then change_tx will send if the change is not `None`

libs/access-control/Cargo.toml
```diff
@@ -8,7 +8,11 @@ actix-http.workspace = true
 app-error.workspace = true
 anyhow.workspace = true
 async-trait.workspace = true
-casbin = { version = "2.2.0", features = ["cached", "runtime-tokio", "incremental"] }
+casbin = { version = "2.2.0", features = [
```
appflowy, line 11: we can try to use 2.5.0 version

libs/access-control/src/noops/collab.rs
```diff
@@ -0,0 +1,95 @@
...
+use async_trait::async_trait;
+use database_entity::dto::AFAccessLevel;
+
+use crate::{
+  act::Action,
+  collab::{CollabAccessControl, RealtimeAccessControl},
+};
+
+#[derive(Clone)]
+pub struct CollabAccessControlImpl {}
```
appflowy, line 11: can use use `pub struct CollabAccessControlImpl;`

libs/access-control/src/noops/collab.rs
```diff
@@ -0,0 +1,95 @@
...
+    Ok(())
+  }
+
+  async fn remove_access_level(&self, _uid: &i64, _oid: &str) -> Result<(), AppError> {
+    Ok(())
+  }
+}
+
+#[derive(Clone)]
+pub struct RealtimeCollabAccessControlImpl {}
```
appflowy, line 62: save as abve

libs/access-control/src/noops/workspace.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+impl Default for WorkspaceAccessControlImpl {
+  fn default() -> Self {
+    Self::new()
+  }
+}
+
+#[async_trait]
+impl WorkspaceAccessControl for WorkspaceAccessControlImpl {
```
appflowy, line 25: Not sure we should provide all default implement in AccessControl traits. Then we don't need to provide the default implementation its `impl`

libs/access-control/Cargo.toml
khorshuheng: Just tried using the 2.5.0 version - there are some changes in function signature. Will do the upgrade in another PR instead and keep 2.2.0 for this PR.

PR #862, opened 2024-10-06
chore: using copy to improve write speed of batch collab api endpoint

appflowy:
1. use rayon to do decompression and data validation
2. using copy instead of insert one by one to postgres in the background queue

### DIFF
--- a/services/appflowy-collaborate/src/collab/storage.rs
+++ b/services/appflowy-collaborate/src/collab/storage.rs
@@ -229,6 +229,14 @@ where
     let mut transaction = self.cache.pg_pool().begin().await?;
     insert_into_af_collab_bulk_for_user(&mut transaction, uid, workspace_id, &params_list).await?;
     transaction.commit().await?;
+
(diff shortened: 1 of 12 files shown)


services/appflowy-collaborate/src/collab/storage.rs
```diff
@@ -270,23 +283,23 @@ where
...
   ) -> AppResult<()> {
-    params.validate()?;
-
     self
       .check_write_workspace_permission(workspace_id, uid)
       .await?;
+    for params in &params_list {
+      self
+        .access_control
+        .update_policy(uid, &params.object_id, AFAccessLevel::FullAccess)
```
Horusiath, line 306: If `batch_insert_collabs` should work on batches, should that for loop be a batch as well?

appflowy: yes. it should be. Because shuheng is working on access control. So I don't want to modify the access control. I will leave a todo for myself

PR #866, opened 2024-10-08
Feat/oauth server

speed2exe:
- Allow AppFlowy-Cloud to become oauth server, i.e. gives the ability for other application to login with AppFlowy-Cloud


admin_frontend/src/session.rs
```diff
@@ -62,68 +72,144 @@ impl SessionStorage {
...
 
-impl UserSession {
-  pub fn new(session_id: String, token: GotrueTokenResponse) -> Self {
-    Self { session_id, token }
+impl FromRedisValue for CodeSessionOptional {
+  fn from_redis_value(v: &redis::Value) -> redis::RedisResult<Self> {
+    let bytes = expect_redis_value_data(v)?;
+    match bytes {
+      Some(bytes) => {
+        let session = expect_redis_json_bytes(bytes).unwrap();
```
appflowy, line 117: is it safe to unwrap? Or we can return None if `expect_redis_json_bytes` returns error

speed2exe: it's safe in the sense that the server is the one serializing and de serializing, and only will panic if there is some corruption. but nonetheless, I'll use the `?` to return the error

appflowy:
we need to write some tests for the outh server.

speed2exe:
I've added some tests for oauth2 server

almereyda:
Is it also planned for AppFlowy to become a generic OIDC client, in so other authentication options can be plugged?

- https://github.com/AppFlowy-IO/AppFlowy-Cloud/issues/621
- https://github.com/AppFlowy-IO/AppFlowy/issues/5534

speed2exe:
@almereyda AFAIK different oauth server may have differences in oauth implementation. See: https://github.com/supabase/auth/tree/master/internal/api/provider thus attempting providing a generic oauth2 client is probably not a good idea.

Alternative solution that you can consider: 
- Email and Password 
This is created during registration, and can be changed in the web admin
- Login using link in the email
During login in, fill in only Email without password. a sign in url will be sent to the email
- SAML Assertion, see:
https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/doc/OKTA_SAML.md

PR #877, opened 2024-10-14
feat: support appflowy web editing for document

khorshuheng:
Add API endpoint to receive updates from AppFlowy Web and apply the changes to the collab stored in-memory / on the database. As of now AppFlowy Web does not need to be updated until the browser is refreshed, hence there is no need to send missing updates back to the web client.


appflowy:
there is 10 MB payload limit for each API endpoint.However, this should be sufficient for your case. If it isn't, you may need to update the NGINX configuration.

PR #882, opened 2024-10-14
feat: CRUD interface for custom namespace

speed2exe:
Functionalities
- Given workspace_id, get all published views info (with some actual doc info).
- Given workspace_id and view_id, sets default publish view for the workspace
- Given workspace_id, get default publish view_info
- Given workspace namespace, get publish view_info and publish view meta


libs/client-api/src/http_publish.rs
```diff
@@ -192,6 +249,31 @@ impl Client {
...
+    );
+
+    let resp = self
+      .cloud_client
+      .get(&url)
+      .send()
+      .await?
+      .error_for_status()?;
+
+    AppResponse::<PublishInfoMeta<T>>::from_response(resp)
```
appflowy, line 272: Each api endpoint should call `log_request_id`.

libs/database-entity/src/dto.rs
```diff
@@ -397,11 +408,19 @@ pub struct AFCollabMember {
   pub permission: AFPermission,
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize)]
 pub struct PublishInfo {
   pub namespace: Option<String>,
   pub publish_name: String,
   pub view_id: Uuid,
+  pub publisher_email: String,
```
appflowy, line 416: You should provide default value. Otherwise, new frontend client use old version appflowy-cloud will get serde error

speed2exe: In what scenario will there be new frontend client using old appflowy-cloud? I'd assumed that we will be upgrading appflowy-cloud first before the releasing new appflowy flutter

libs/client-api/src/http_publish.rs
speed2exe: will add it in

libs/database-entity/src/dto.rs
khorshuheng: For self hosting user, mostly. For us there should be minimal impact.

speed2exe: added default values

PR #885, opened 2024-10-15
feat: additional access control config

khorshuheng:
Add additional access control configuration such that it is possible to enforce access control only for a subset of the endpoints.

Also added missing permission checks for workspace member related CRUD, and update access control when an access request is approved by the workspace owner.

Simplify NotEnoughPermission error type, such that it doesn't require a parameter. From the caller's point of view, the current error message has not much value because the caller is already aware of the user's identify and the action that the user is trying to perform.


speed2exe:
LGTM

PR #913, opened 2024-10-21
chore: publish metadata wrapper

speed2exe:
- Wraps the original publish metadata response with `AppResponse` 
- Retains but deprecate the old endpoint `/published/{publish_namespace}/{publish_name}`
- Client Api now uses `/v1/published/{publish_namespace}/{publish_name}`

### DIFF
--- a/libs/client-api/src/http_publish.rs
+++ b/libs/client-api/src/http_publish.rs
@@ -292,7 +292,7 @@ impl Client {
     publish_name: &str,
   ) -> Result<T, AppResponseError>
   where
-    T: serde::de::DeserializeOwned,
+    T: serde::de::DeserializeOwned + 'static,
   {
     tracing::debug!(
       "get_published_collab: {} {}",
@@ -300,7 +300,7 @@ impl Client {
       publish_name
     );
     let url = format!(
-      "{}/api/workspace/published/{}/{}",
+      "{}/v1/api/workspace/published/{}/{}",
       self.base_url, publish_namespace, publish_name
     );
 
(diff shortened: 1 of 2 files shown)


libs/client-api/src/http_publish.rs
```diff
@@ -292,15 +292,15 @@ impl Client {
...
+    T: serde::de::DeserializeOwned + 'static,
   {
     tracing::debug!(
       "get_published_collab: {} {}",
       publish_namespace,
       publish_name
     );
     let url = format!(
-      "{}/api/workspace/published/{}/{}",
+      "{}/v1/api/workspace/published/{}/{}",
```
khorshuheng, line 303: As we are using /api for our nginx config for calls to appflowy cloud, we probably should still keep the `api` prefix instead.

src/api/workspace.rs
```diff
@@ -1230,6 +1235,18 @@ async fn get_published_collab_handler(
...
+  path_param: web::Path<(String, String)>,
+  state: Data<AppState>,
+) -> Result<Json<AppResponse<serde_json::Value>>> {
+  let (workspace_namespace, publish_name) = path_param.into_inner();
+  let metadata = state
+    .published_collab_store
+    .get_collab_metadata(&workspace_namespace, &publish_name)
+    .await?;
+  Ok(Json(AppResponse::Ok().with_data(metadata)))
+}
```
khorshuheng, line 1248: In this case, the response is in the form of:
```
{
   "data": {
     "ancestor_views": [...]
   }
}
```
Is this what the front ends expects? Or should it be:

{
  "data": {
    "meta": {
      "ancestor_views": [...]
    }
  }
}
```

libs/client-api/src/http_publish.rs
speed2exe: `v1` as what you mentioned can potentially conflict with namespace. This also follows the convention of `v1` the previous endpoints.

src/api/workspace.rs
speed2exe: Yes. I have confirmed with @qinluhe on this.

libs/client-api/src/http_publish.rs
speed2exe: oh sorry, I was mistaken, you were right. I've made the changes

khorshuheng:
lgtm, @qinluhe do check again if the new response format is expected.

src/api/workspace.rs
qinluhe: LGTM

PR #914, opened 2024-10-21
chore: access control interface to return error instead of boolean

khorshuheng:
Currently, the access control trait will return true if the subject has access to the object, and false otherwise. This makes it possible to enforce the behavior of the implementation: it is very clear that the app error means unexpected behaviour, and the boolean return value indicate whether the user has access or not.

However, on the caller's end - it became impossible to enforce the error type to return if the user does not have the access. It is very possible for the caller to accidentally return a different error code, which will result in inconsistency. There's also additional handling required for every call, to return permission error if the result is false.

It is possible to enforce the returned error code by having a middleware that will be used by every single endpoints that requires access control. However, our long term plan is to stop relying on middleware, and expect the endpoints to use the access control interface directly. This is due to three reasons:
1. middleware on the global level introduce a layer of abstraction which the developer can accidentally missed
2. Some endpoints requires additional context to enforce access control, which might not be available on the middleware layer.
3. Some endpoints might provide a fallback response in the event when the user does not have sufficient permission.

Therefore, the alternative is to return NotEnoughPermission error in the event where the user does not have access, which will simplify the usage of the access control trait.

The downside, is that it is possible for an implementation to return a different error type than expected when a user does not have access. However, since it is not expected for us to have many implementations, aside from casbin and noops, this should be manageable by documenting the trait properly.

### DIFF
--- a/libs/access-control/src/casbin/access.rs
+++ b/libs/access-control/src/casbin/access.rs
@@ -99,7 +99,7 @@ impl AccessControl {
     uid: &i64,
     obj: ObjectType<'_>,
     act: ActionVariant<'_>,
-  ) -> Result<bool, AppError> {
+  ) -> Result<(), AppError> {
     self
       .enforcer
       .enforce_policy(workspace_id, uid, obj, act)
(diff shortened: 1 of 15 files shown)


speed2exe (approved):
LGTM

Issue #839, opened 2024-09-24
[Bug] Supabase Auth / gotrue migration 20240612123726 fails on Postgres container without postgres role


almereyda:
**Describe the bug**

The Supabase Auth / gotrue migration https://github.com/supabase/auth/blob/master/migrations/20240612123726_enable_rls_update_grants.up.sql hard codes a `postgres` role, which is non-existent on Postgres containers which set the `POSTGRES_USER` variable.

**To Reproduce**

Steps used to reproduce the behavior in with #622:

1. `docker compose pull`
2. `docker compose rm -sf`
3. `docker compose build`
4. `docker compose up -d`
5. `docker compose logs gotrue`

**Expected behavior**

All migrations run.

**Evidence**

```
[...]

PR #918, opened 2024-10-22
feat: simplify env settings

ThanatosDi:
- simplify env settings, reuse varable in .env file
- ignore supabase user .sql file init
- create supabase user init shell script, make custom supabase username and password possible


speed2exe:
@ThanatosDi For self-hosters, I suppose they will have to run the shell script `migrations/before/supabase_auth.sh` after docker compose commands? Do you mind adding some documentation?

khorshuheng:
> @ThanatosDi For self-hosters, I suppose they will have to run the shell script `migrations/before/supabase_auth.sh` after docker compose commands? Do you mind adding some documentation in https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/doc/DEPLOYMENT.md?

There is no need to run the sh script manually - docker-entrypoint init will execute sh script as well as sql script.

khorshuheng:
One issue right now, is that changing the POSTGRES USER variable will cause this issue:
https://github.com/AppFlowy-IO/AppFlowy-Cloud/issues/839

khorshuheng:
@ThanatosDi Apart from the issue i mentioned above, this looks ok to me. There's also no need to keep a backup in the form of `.bak` - if it is necessary to revert we can check the commit history.

We will need to spend some time to see if it is possible for us to solve #839 - it is certainly possible to fix this for the new user, but i still need to find a way to patch gotrue service without breaking existing installation.

If we can't find a fix in the near future, then it is perhaps better to make POSTGRES_USER non configurable for now.

deploy.env
```diff
@@ -1,23 +1,39 @@
...
+REDIS_PORT=6379
+
+# Gotrue Settings
+GOTRUE_HOST=gotrue
+GOTRUE_PORT=9999
 
 # AppFlowy Cloud
 ## URL that connects to the gotrue docker container
-APPFLOWY_GOTRUE_BASE_URL=http://gotrue:9999
+APPFLOWY_GOTRUE_BASE_URL=http://${GOTRUE_HOST}:${GOTRUE_PORT}
```
khorshuheng, line 25: There's actually no reason to change gotrue host, since we are not going to use the publicly accessible endpoint. We can keep it as gotrue:9999.

This applies to Redis as well.

khorshuheng:
If Redis needs to be made configurable, then we will need to change this as well: https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/docker-compose.yml#L94

ThanatosDi:
> > @ThanatosDi For self-hosters, I suppose they will have to run the shell script `migrations/before/supabase_auth.sh` after docker compose commands? Do you mind adding some documentation in https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/doc/DEPLOYMENT.md?
> 
> There is no need to run the sh script manually - docker-entrypoint init will execute sh script as well as sql script.

yes, docker-entrypoint init will auto execute shell script when database is empty.

> If Redis needs to be made configurable, then we will need to change this as well: https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/docker-compose.yml#L94

ok, I will change docker-compose.yml let redis URI use .env variable

khorshuheng (approved):

PR #919, opened 2024-10-22
feat: add publish info to template

khorshuheng:
Add publish info to get template endpoint's response, to be used by the web frontend.


speed2exe (approved):
LGTM

PR #921, opened 2024-10-23
feat: add publish info to list templates

khorshuheng:
Add publish info to list templates endpoint. It is assumed that published info can always be found for a given template/view id.

### DIFF
--- a/libs/database-entity/src/dto.rs
+++ b/libs/database-entity/src/dto.rs
@@ -1344,22 +1344,22 @@ pub struct TemplateWithPublishInfo {
   pub publish_info: PublishInfo,
 }
 
-impl TemplateWithPublishInfo {
-  pub fn from_template_and_publish_info(template: &Template, publish_info: &PublishInfo) -> Self {
+impl From<(Template, PublishInfo)> for TemplateWithPublishInfo {
+  fn from((template, publish_info): (Template, PublishInfo)) -> Self {
     Self {
       view_id: template.view_id,
       created_at: template.created_at,
       last_updated_at: template.last_updated_at,
-      name: template.name.clone(),
-      description: template.description.clone(),
-      about: template.about.clone(),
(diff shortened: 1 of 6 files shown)


libs/database-entity/src/dto.rs
```diff
@@ -1378,9 +1378,45 @@ pub struct TemplateMinimal {
...
+  pub creator: TemplateCreatorMinimal,
+  pub categories: Vec<TemplateCategoryMinimal>,
+  pub is_new_template: bool,
+  pub is_featured: bool,
+  pub publish_info: PublishInfo,
+}
+
+impl TemplateMinimalWithPublishInfo {
+  pub fn from_template_and_publish_info(
+    template: &TemplateMinimal,
```
speed2exe, line 1398: Can the params be owned instead? Using a ref to `Template`Minimal` and `PublishInfo` seem to have a lot of cloning. I'm also wondering why not just impl the `std::convert::From` for `(TemplateMinimal, PublishInfo)` tuple type.

src/biz/template/ops.rs
```diff
@@ -245,7 +245,28 @@ pub async fn get_templates(
...
+    select_published_collab_info_for_view_ids(pg_pool, &view_ids).await?;
+  let publish_info_map = publish_info_for_views
+    .into_iter()
+    .map(|info| (info.view_id, info))
+    .collect::<HashMap<Uuid, PublishInfo>>();
+  let templates_with_publish_info: Vec<TemplateMinimalWithPublishInfo> = templates
+    .into_iter()
+    .filter_map(|template| {
+      publish_info_map
+        .get(&template.view_id)
```
speed2exe, line 259: Since `publish_info_map` is not being used afterwards, you could use 'remove' instead of `get` to own the `PublishInfo`, avoiding the need for 'clone()'

speed2exe:
LGTM

PR #923, opened 2024-10-23
fix: allow gotrue service to work without postgres user

khorshuheng:
This addresses https://github.com/AppFlowy-IO/AppFlowy-Cloud/issues/839 . For most deployment setup of appflowy cloud, there is no strong reason of having row level security, since the database host is usually not going to be publicly accessible.


speed2exe (approved):
LGTM

Issue #907, opened 2024-10-19
[Bug] Do not show credentials in the logs !


chevdor:
**Describe the bug**
No credential should ever be shown in the logs...
See https://github.com/AppFlowy-IO/AppFlowy/issues/6583 for an example.

**To Reproduce**
1. look at the startup logs

**Expected behavior**
No credential is **ever** shown in the logs

**Screenshots**
<img width="1568" alt="image" src="https://github.com/user-attachments/assets/70f42021-8748-4df9-b6b6-34de4571bb64">

**Desktop (please complete the following information):**
n/a

**Additional context**
n/a

wolfspyre:
yeah. this **MUST** be fixed. 
additionally; I'd encourage tests be put in place to inspect the emitted logs for sensitive data... 

(this stuff has a nasty habit of getting re-enabled from dev pr's.... having a test to make sure that no verboselog happens to shout your `hunter2` to the wide-wide-world-of-logs.... will make it a lot easier to ensure it doesnt happen again.

PR #925, opened 2024-10-23
fix: masked postgres password in logs

khorshuheng:
Fix https://github.com/AppFlowy-IO/AppFlowy-Cloud/issues/907 .

### DIFF
--- a/services/appflowy-history/src/config.rs
+++ b/services/appflowy-history/src/config.rs
@@ -3,7 +3,7 @@ use collab_stream::client::CONTROL_STREAM_KEY;
 use infra::env_util::get_env_var;
 use serde::Deserialize;
 use sqlx::postgres::{PgConnectOptions, PgSslMode};
-use std::str::FromStr;
+use std::{fmt::Display, str::FromStr};
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


speed2exe:
Why not just use the `secrecy::Secret` like how `redis_uri` is defined (`pub redis_uri: Secret<String>`) in `src/config/config.rs:21`?

khorshuheng:
> Why not just use the `secrecy::Secret` like how `redis_uri` is defined (`pub redis_uri: Secret<String>`) in `src/config/config.rs:21`?

It's mainly because of `PgConnectOption`:
```
pub struct DatabaseSetting {
  pub pg_conn_opts: PgConnectOptions,
```

If we want to use secrecy::Secret, then we should update this struct to have the database url instead of pg_conn_opts as part of the field, and create a function to return pg_conn_opts based on the database url instead. Would that be preferable?

khorshuheng:
There's also another issue - we are passing database url instead of host, port and password. So if we use Secret, the entire database url will be Redacted, and in which case, we are better off not printing the configuration for pg connect options at all.

speed2exe (approved):

PR #931, opened 2024-10-24
feat: patching of publish name

speed2exe:
- Allow user to edit published view specifically `publish_name`
- API is designed for batch operation and may support other patching operation (aside from `publish_name`) in the future.


khorshuheng (approved):
Lgtm

PR #938, opened 2024-10-26
chore: add limit for importing zip file

appflowy:


libs/client-api/src/native/http_native.rs
```diff
@@ -362,6 +366,79 @@ impl Client {
...
+      .body(stream_body)
+      .send()
+      .await?;
+
+    if !upload_resp.status().is_success() {
+      error!("File upload failed: {:?}", upload_resp);
+      return Err(AppError::S3ResponseError("Cannot upload file to S3".to_string()).into());
+    }
+
+    Ok(())
```
qinluhe, line 439: I think we should return the file url or file id after uploading the file, and then pass this file url or fileid when creating import. The file name is not accurate.

appflowy: call create import first and then call upload_import_file.

appflowy: File name will be the workspace name.

Issue #934, opened 2024-10-25
[Bug] supabase_auth_admin user not found


ThanatosDi:
**Describe the bug**

PR #918 can custom setting supabase username, but miss check migrations, will let that init failed.
some migrations need origin username setting "supabase_auth_admin".

**Fix**

1. set supabase username hard code, still use "supabase_auth_admin", and custom username is not allowed.
2. change some migrations from .sql to shell script, that can get value from env variable

Any suggestions for these two solutions?

@speed2exe , @khorshuheng

khorshuheng:
Option 2 would not be possible, because the migration SQL files are used by the sqlx library, and as far as I know, does not support shell script nor variable substitution.

I would suggest going with option 1 for now. Meanwhile, I will explore other options.

PR #939, opened 2024-10-26
refactor: remove supabase username variable and force set to "supabase_auth_admin"

ThanatosDi:
#934

### DIFF
--- a/deploy.env
+++ b/deploy.env
@@ -9,7 +9,6 @@ POSTGRES_PORT=5432
 POSTGRES_DB=postgres
 
 # Supabase user settings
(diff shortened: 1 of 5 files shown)


deploy.env
```diff
@@ -9,7 +9,7 @@ POSTGRES_PORT=5432
 POSTGRES_DB=postgres
 
 # Supabase user settings
-SUPABASE_USER=supabase_auth_admin
+# SUPABASE_USER=supabase_auth_admin #934 force set username supabase_auth_admin
```
khorshuheng, line 12: We can remove this entirely, rather than commenting it out.

PR #940, opened 2024-10-28
feat: delete default view for workspace

speed2exe:
- Add http api for removing default view id for a workspace

### DIFF
--- a/src/api/workspace.rs
+++ b/src/api/workspace.rs
@@ -1151,7 +1151,6 @@ async fn put_workspace_default_published_view_handler(
   let new_default_pub_view_id = payload.into_inner().view_id;
   biz::workspace::publish::set_workspace_default_publish_view(
     &state.pg_pool,
-    &user_uuid,
     &workspace_id,
     &new_default_pub_view_id,
   )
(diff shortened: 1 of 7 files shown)


src/api/workspace.rs
```diff
@@ -1158,6 +1159,25 @@ async fn put_workspace_default_published_view_handler(
...
 
+async fn delete_workspace_default_published_view_handler(
+  user_uuid: UserUuid,
+  workspace_id: web::Path<Uuid>,
+  state: Data<AppState>,
+) -> Result<Json<AppResponse<()>>> {
+  let uid = state.user_cache.get_user_uid(&user_uuid).await?;
+  state
+    .workspace_access_control
+    .enforce_action(&uid, &workspace_id.to_string(), Action::Write)
```
khorshuheng, line 1170: If only workspace owner should be allowed to delete default published view, you can use enfore_role instead.

src/biz/workspace/publish.rs
```diff
@@ -113,6 +113,16 @@ pub async fn set_workspace_default_publish_view(
   Ok(())
 }
 
+pub async fn unset_workspace_default_publish_view(
+  pg_pool: &PgPool,
+  user_uuid: &Uuid,
+  workspace_id: &Uuid,
+) -> Result<(), AppError> {
+  check_workspace_owner(pg_pool, user_uuid, workspace_id).await?;
```
khorshuheng, line 121: If we use enforce role: Owner, we don't need to check workspace owner again.

src/api/workspace.rs
speed2exe: done

src/biz/workspace/publish.rs
speed2exe: Removed. In addition, I have remove 'check_workspace_owner' and use `enforce_role` on the caller instead for other operation that requires `check_workspace_owner`

khorshuheng (approved):
lgtm

PR #944, opened 2024-10-28
feat: handle duplicated publish names for a workspace

speed2exe:
- If a collab is already publish under a `view_id` and user tries to publish another `view_id` with same `publish_name` as the previous, an error `PublishNameAlreadyExists = 1050` will be returned.
- If a collab is already published, user cannot change the `publish_name` to one that already exists for the workspace, error `PublishNameAlreadyExists = 1050` will be returned


khorshuheng (approved):
lgtm

LucasXu0:
If the publish name contains '/', '&', or '?', what's the error code returned?

speed2exe:
@LucasXu0 It's `InvalidRequest = 1008`.

PR #962, opened 2024-11-04
feat: api for space and default collab creation

khorshuheng:
Add API endpoint so that AppFlowy Web can create new page. Only support document at the moment.


src/biz/workspace/page_view.rs
```diff
@@ -32,15 +35,170 @@ use crate::biz::collab::{
...
+}
+
+pub async fn create_page(
+  pg_pool: &PgPool,
+  collab_storage: &CollabAccessControlStorage,
+  uid: i64,
+  workspace_id: Uuid,
+  parent_view_id: &str,
+  view_layout: &ViewLayout,
+) -> Result<Page, AppResponseError> {
```
speed2exe, line 50: as much as possible, we should use `AppResponseError` only at the `api` crate, business logic in `biz` should use `AppError` instead

tests/workspace/page_view.rs
```diff
@@ -39,3 +42,53 @@ async fn get_page_view() {
...
+  let page = c
+    .create_workspace_page_view(
+      workspace_id,
+      &CreatePageParams {
+        parent_view_id: general_space.view_id.clone(),
+        layout: ViewLayout::Document,
+      },
+    )
+    .await
+    .unwrap();
```
speed2exe, line 70: I have experienced flaky test in template duplication test cases, which use the same mechanism to broadcast all updates to client. I solved at by waiting for 1 second before the next api call. If you experienced flaky test scenario in future test runs, can consider to use this technique.

PR #965, opened 2024-11-05
feat: api for moving page to trash

khorshuheng:
Add API to move page to trash folder. The page will be removed from favorite before moving to trash, similar to the desktop app behaviour.

### DIFF
--- a/libs/client-api/src/http_view.rs
+++ b/libs/client-api/src/http_view.rs
@@ -1,5 +1,6 @@
 use client_api_entity::workspace_dto::{CreatePageParams, Page, PageCollab};
 use reqwest::Method;
+use serde_json::json;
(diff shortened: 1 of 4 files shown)


libs/client-api/src/http_view.rs
```diff
@@ -21,6 +21,23 @@ impl Client {
...
+    &self,
+    workspace_id: Uuid,
+    view_id: String,
+  ) -> Result<(), AppResponseError> {
+    let url = format!(
+      "{}/api/workspace/{}/page-view/{}/move-to-trash",
+      self.base_url, workspace_id, view_id
+    );
+    let resp = self
+      .http_client_with_auth(Method::POST, &url)
```
speed2exe, line 34: maybe we should use `Method::DELETE` instead? then you dont need the `move-to-trash` in the path

PR #967, opened 2024-11-06
feat: API for updating page icon, name, and extra

khorshuheng:
This PR add API to update page icon, name, and extra. Frontend application is expected to supply None for icon if the intention is to remove the icon, and some string values otherwise.


speed2exe (approved):
LGTM

PR #974, opened 2024-11-08
feat: api for restore all pages from trash

khorshuheng:
API endpoint to support restore all pages from trash.


speed2exe (approved):
LGTM

PR #990, opened 2024-11-13
chore: split document content into multiple embeddings

Horusiath:
This PR splits indexed document content into multiple embeddings of ~2000 tokens in size.


services/appflowy-collaborate/src/indexer/document_indexer.rs
```diff
@@ -42,16 +45,42 @@ impl Indexer for DocumentIndexer {
...
-          fragment_id: object_id.clone(),
-          object_id: object_id.clone(),
-          collab_type: CollabType::Document,
-          content_type: EmbeddingContentType::PlainText,
-          content,
-          embedding: None,
+        let mut result = Vec::with_capacity(1 + content.len() / Self::DOC_CONTENT_SPLIT);
+
+        let mut slice = content.as_str();
+        while slice.len() > Self::DOC_CONTENT_SPLIT {
```
appflowy, line 51: I can handle content splitting directly in the AppFlowy-AI service, eliminating the need to split it in AppFlowy Cloud using DOC_CONTENT_SPLIT.

libs/database/src/index/collab_embeddings_ops.rs
```diff
@@ -77,34 +77,35 @@ pub async fn upsert_collab_embeddings(
     .await?;
   }
 
-  for r in records {
-    sqlx::query(
-      r#"INSERT INTO af_collab_embeddings (fragment_id, oid, partition_key, content_type, content, embedding, indexed_at)
+  if !records.is_empty() {
+    // replace existing collab embeddings
+    remove_collab_embeddings(tx, &records[0].object_id).await?;
+    for r in records {
```
appflowy, line 83: Should we use [Bulk Inserts](https://www.postgresql.org/docs/current/functions-array.html#FUNCTIONS-ARRAY-TABLE) here?

services/appflowy-collaborate/src/indexer/document_indexer.rs
Horusiath: From what I've noticed, text splitting in AppFlowy-AI service is not what we need: It uses either \n\n for text splitting (which roughly means splitting by paragraph) or splitting by code formatting scopes. Neither of these is good for our use case - we were explicitly splitting by block in the past and it turned out to be too expensive, which is why we are at the current state.

libs/database/src/index/collab_embeddings_ops.rs
Horusiath: Yes, we could. We could also do it in another PR.

PR #995, opened 2024-11-15
fix: byte index 8000 is not a char boundary

appflowy:


Horusiath (approved):

services/appflowy-collaborate/src/indexer/document_indexer.rs
```diff
@@ -141,3 +112,211 @@ impl Indexer for DocumentIndexer {
...
+  object_id: String,
+  content: String,
+  collab_type: CollabType,
+  max_content_len: usize,
+) -> Result<Vec<AFCollabEmbeddingParams>, AppError> {
+  let mut result = Vec::with_capacity(1 + content.len() / max_content_len);
+
+  let mut start = 0;
+  let mut current_len = 0;
+  let graphemes: Vec<&str> = content.graphemes(true).collect();
```
Horusiath, line 126: In 90% of the cases you won't ever need to split the string at all, so allocating array of graphemes seems wasteful.

services/appflowy-collaborate/src/indexer/document_indexer.rs
```diff
@@ -141,3 +112,211 @@ impl Indexer for DocumentIndexer {
...
+
+  for (i, grapheme) in graphemes.iter().enumerate() {
+    current_len += grapheme.len();
+    if current_len > max_content_len {
+      let fragment = graphemes[start..i].concat();
```
Horusiath, line 131: You can also do:

```suggestion
      let fragment = content[start..current_len].to_string();
```
where `start` is the index of previous split or 0. This way you don't need to concat 8000 characters for every split, just do a memcopy. 

With this you also don't need to enumerate, since we don't care about grapheme positions, only about their length.

PR #998, opened 2024-11-15
feat: allow underscore in publish url

speed2exe:


src/biz/workspace/publish.rs
```diff
@@ -74,7 +74,7 @@ fn check_collab_publish_name(publish_name: &str) -> Result<(), AppError> {
 
   // Only contain alphanumeric characters and hyphens
   for c in publish_name.chars() {
-    if !c.is_alphanumeric() && c != '-' {
+    if !c.is_alphanumeric() && c != '-' && c != '_' {
```
LucasXu0, line 77: should we use regex to handle it?

speed2exe: If we there is more complicated pattern (like email) that we want to enforce, I think it would be worth to add it in. For now, this is good enough.

PR #1000, opened 2024-11-15
chore: store pending collab writes in memory

Horusiath:
This PR replaced Redis pending collab writes queue with in-process channel. There are few reasons for that:
1. If we want to scale out number of web service instances in the future they shouldn't override collab request: which is what the current implementation would do.
2. It's much faster this way as there are no extra network hops.
3. We don't need strong persistence of states: since server is not the source of truth, we don't need to worry what will happen when the collab is not persisted. If we ie. restart server, when a client connects, it will simply resend collab's missing updates.


libs/client-websocket/src/error.rs
```diff
@@ -59,7 +59,7 @@ pub enum Error {
   #[error("URL error: {0}")]
   Url(#[from] UrlError),
   #[error("HTTP error: {}", .0.status())]
-  Http(Response<Option<Vec<u8>>>),
+  Http(Box<Response<Option<Vec<u8>>>>),
```
Horusiath, line 62: More recent versions of clippy reminds that with this structure inlined directly, the Error enum struct is over 136B in size.

appflowy:
When a user wants to confirm whether the server has saved their data, do we need to query the database? Will it take too long for the user to notice that the data has been saved?

Horusiath:
On the current main branch there's no such confirmation. In this branch, if we use save collab with `write_immediate` option it will try to write collab down immediately (without using a queue) and return when it's completed.

I was thinking about adding something like `Option<tokio::sync::oneshot::Sender<Result<()>>>` as a method param that would be confirmed when persister loop actually manages to save the collab.

appflowy:
Before implementing a queue, we encountered a pool timeout issue. For instance, simultaneous writes or other API endpoints occupying pool connections could lead to pool timeouts. If pool timeouts occur more frequently, how can we address this issue?

services/appflowy-collaborate/src/collab/storage.rs
```diff
@@ -75,6 +89,103 @@ where
     }
   }
 
+  const PENDING_WRITE_BUF_CAPACITY: usize = 100;
+  async fn periodic_write_task(cache: CollabCache, mut reader: Receiver<PendingCollabWrite>) {
+    let mut buf = Vec::with_capacity(Self::PENDING_WRITE_BUF_CAPACITY);
```
appflowy, line 94: Is it possible for Self::persist to be called very frequently?

services/appflowy-collaborate/src/collab/storage.rs
```diff
@@ -75,6 +89,103 @@ where
     }
   }
 
+  const PENDING_WRITE_BUF_CAPACITY: usize = 100;
```
appflowy, line 92: Maybe we can use env here. It's easy to adjust the capacity

services/appflowy-collaborate/src/collab/storage.rs
```diff
@@ -57,15 +74,12 @@ where
...
-    redis_conn_manager: RedisConnectionManager,
-    metrics: Arc<CollabMetrics>,
   ) -> Self {
-    let queue = Arc::new(StorageQueue::new_with_metrics(
-      cache.clone(),
-      redis_conn_manager,
-      REDIS_PENDING_WRITE_QUEUE,
-      Some(metrics),
-    ));
+    let (queue, reader) = channel(1000);
```
appflowy, line 78: Should we have metrics for the queue?

services/appflowy-collaborate/src/collab/storage.rs
```diff
@@ -75,6 +89,103 @@ where
...
+        .await
+      {
+        sqlx::query(&format!("ROLLBACK TO SAVEPOINT {}", savepoint_name))
+          .execute(transaction.deref_mut())
+          .await?;
+      }
+    }
+
+    // Commit the transaction to finalize all writes
+    match tokio::time::timeout(Duration::from_secs(10), transaction.commit()).await {
```
appflowy, line 144: Assuming we have PENDING_WRITE_BUF_CAPACITY set for each write, I’m not sure if 10 seconds is sufficient.

Horusiath: This part is literally copy-pasted from current StorageQueue.

services/appflowy-collaborate/src/collab/storage.rs
Horusiath: According to documentation, `reader.recv_many` used in a loop below will await until at least one element has been passed to a reader or the sender channel has been closed.

services/appflowy-collaborate/src/collab/storage.rs
Horusiath: I didn't put it into configurable value since it's also hardcoded in the current implementation: see `consume_pending_write(&pending_write_set, 20, 5).await;` in the deleted `queue.rs` file.

services/appflowy-collaborate/src/collab/storage.rs
appflowy: I believe they are different. The StorageQueue processes 20 collabs at a time, whereas, in some cases, you process 100 collabs at once.
```rust
 let chunk_keys = consume_pending_write(&pending_write_set, 20, 5).await;
      if chunk_keys.is_empty() {
        continue;
      }

      for keys in chunk_keys {
        trace!(
          "start writing {} pending collaboration data to disk",
          keys.len()
        );
        let cloned_collab_cache = collab_cache.clone();
        let mut cloned_connection_manager = connection_manager.clone();
        let cloned_total_write_count = total_write_count.clone();
        let cloned_total_success_write_count = success_write_count.clone();

        if let Ok(metas) = get_pending_meta(&keys, &mut cloned_connection_manager).await {
          if metas.is_empty() {
            error!("the pending write keys is not empty, but metas is empty");
            return;
          }

          match retry_write_pending_to_disk(&cloned_collab_cache, metas).await {
            Ok(success_result) => {
              #[cfg(debug_assertions)]
              tracing::info!("success write pending: {:?}", keys,);

              trace!("{:?}", success_result);
              cloned_total_write_count.fetch_add(
                success_result.expected as i64,
                std::sync::atomic::Ordering::Relaxed,
              );
              cloned_total_success_write_count.fetch_add(
                success_result.success as i64,
                std::sync::atomic::Ordering::Relaxed,
              );
            },
            Err(err) => error!("{:?}", err),
          }
          // Remove pending metadata from Redis even if some records fail to write to disk after retries.
          // Records that fail repeatedly are considered potentially corrupt or invalid.
          let _ = remove_pending_meta(&keys, &mut cloned_connection_manager).await;
        }
      }
```

services/appflowy-collaborate/src/collab/storage.rs
appflowy: I think you might have misunderstood what the source code does—it writes 20 collabs at a time.

services/appflowy-collaborate/src/collab/storage.rs
Horusiath: I've set it to the same value as on main (20 collabs at a time).

appflowy: OK

PR #1005, opened 2024-11-18
feat: allow create page API to accept a view name

khorshuheng:
Add an optional parameter for create page API, so that the caller can provide the view name.

### DIFF
--- a/libs/shared-entity/src/dto/workspace_dto.rs
+++ b/libs/shared-entity/src/dto/workspace_dto.rs
@@ -132,6 +132,7 @@ pub struct Page {
 pub struct CreatePageParams {
   pub parent_view_id: String,
   pub layout: ViewLayout,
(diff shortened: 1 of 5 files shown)


libs/shared-entity/src/dto/workspace_dto.rs
```diff
@@ -132,6 +132,7 @@ pub struct Page {
 pub struct CreatePageParams {
   pub parent_view_id: String,
   pub layout: ViewLayout,
+  pub name: Option<String>,
```
appflowy, line 135: should add set the serde(default) ? For example, old client connect to latest server might have serde issue

khorshuheng: This API is not being used at the moment, so backward compatibility is not an issue.

PR #1010, opened 2024-11-19
fix: revert changes to casbin matcher

khorshuheng:
Casbin matcher was changed in https://github.com/AppFlowy-IO/AppFlowy-Cloud/commit/d0c212ff10ecf758764bf209fa91db0f5a769dbf to support collab group. However, doing so seems to negatively impact the performance of the policy evaluation.

### DIFF
--- a/libs/access-control/src/casbin/access.rs
+++ b/libs/access-control/src/casbin/access.rs
@@ -159,7 +159,7 @@ g = _, _ # grouping rule
 e = some(where (p.eft == allow))
 
 [matchers]
-m = g(r.sub, p.sub) && p.obj == r.obj && (g(p.act, r.act) || cmpRoleOrLevel(r.act, p.act))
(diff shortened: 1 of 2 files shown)


khorshuheng:
> Can you describe the difference in matchers?

Let say we have a user, 1001. If we use r.sub == p.sub, casbin will check if there's policy matching the condition solely based on the user id. For example, to check whether user 1001 can write to a workspace, there need to be a policy correspond to user 1001 that allows the user to write to the workspace.

On the other hand, using g(r.sub, p.sub) relax the condition, which allow us to have a policy that grant everyone in "group:9001" access to a workspace. As long as user 1001 belongs to group 9001, the user can also access the worspace, eventhough there's no matcher that is explicitly associated with user 1001.

What i suspect, is that g(r.sub, p.sub) takes more time to evaluate than r.sub == p.sub, especially when we have a large number of policies.

Horusiath (approved):

PR #1017, opened 2024-11-24
feat: add a polling api for getting database row id updates

speed2exe:
## Add API endpoint to poll for database row update
- Optionally receives `after` as param, if not given, will be defaulted to 1 hour ago.
- Read after Write is not supported, i.e. if database row is updated or inserted, and this api is called right after, it will not fetch the latest update. Update will be eventually be fetched (within 5 minutes).
- Returns only database row id, not the data itself.

### DIFF
--- a/src/biz/collab/ops.rs
+++ b/src/biz/collab/ops.rs
@@ -3,6 +3,8 @@ use std::sync::Arc;
 
 use app_error::AppError;
 use appflowy_collaborate::collab::storage::CollabAccessControlStorage;
+use chrono::DateTime;
+use chrono::Utc;
 use collab::preclude::Collab;
 use collab_database::database::DatabaseBody;
 use collab_database::entity::FieldType;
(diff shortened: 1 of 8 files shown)


src/biz/collab/ops.rs
```diff
@@ -479,6 +483,25 @@ pub async fn list_database_row(
...
 }
 
+pub async fn list_database_row_ids_updated(
+  collab_storage: &CollabAccessControlStorage,
+  pg_pool: &PgPool,
+  workspace_uuid_str: &str,
+  database_uuid_str: &str,
+  after: &DateTime<Utc>,
+) -> Result<Vec<DatabaseRowUpdatedItem>, AppError> {
+  let row_ids = list_database_row_ids(collab_storage, workspace_uuid_str, database_uuid_str)
```
appflowy, line 493: maybe you can collect as &str

speed2exe: I can collect and make `row_ids` into `Vec<&str>`, but the following function `select_last_updated_database_row_ids` requires `&[String]` for `row_ids`, which then will require a clone. Based on what `into_iter` means, it will own the returned values from `list_database_row_ids`, which doesn't seem to require cloning for the strings.

PR #1022, opened 2024-11-26
chore: reduce number of SQL statements executed when inserting/updating collab

Horusiath:


libs/database/src/collab/collab_db_ops.rs
```diff
@@ -55,109 +55,64 @@ pub async fn insert_into_af_collab(
...
-          permission_id,
-          err
-        ))
-      })?;
+  sqlx::query!(
+    r#"
+  INSERT INTO af_collab (oid, blob, len, partition_key, encrypt, owner_uid, workspace_id)
+  VALUES ($1, $2, $3, $4, $5, $6, $7)
+  ON CONFLICT (oid, partition_key) DO UPDATE
+  SET blob = $2, len = $3, encrypt = $5, owner_uid = $6;
```
Horusiath, line 69: You don't need separate insert/update logic - it can be done in one move.

libs/database/src/collab/collab_db_ops.rs
```diff
@@ -55,109 +55,64 @@ pub async fn insert_into_af_collab(
...
-    },
-  }
+  sqlx::query!(
+    r#"
+  INSERT INTO af_collab_member (uid, oid, permission_id)
+  VALUES ($1, $2, (SELECT rp.permission_id 
+    FROM af_role_permissions rp
+    JOIN af_roles ON rp.role_id = af_roles.id
+    WHERE af_roles.name = 'Owner'))
+  ON CONFLICT (uid, oid)
```
Horusiath, line 99: You don't need to separately select then insert af_collab_member permissions. It can be done in one go.

khorshuheng:
@Horusiath Now that we have simplified collab level access to check workspace instead of collab, i will work on the task to remove `insert into af_collab_member` completely.

Horusiath:
@khorshuheng depending on which PR will be merged first, we can change existing migration that created a stored procedure here, or override it with new migration changing it.

khorshuheng:
> @khorshuheng depending on which PR will be merged first, we can change existing migration that created a stored procedure here, or override it with new migration changing it later. In any case I think this is useful, even if it doesn't fix everything yet.

I will wait for your PR first.

PR #1025, opened 2024-11-27
fix: primary field should be unique for new database page

khorshuheng:
The database fields are wrongly set to true for every fields.


speed2exe (commented):
LGTM

PR #1028, opened 2024-11-28
S3 collab cache

Horusiath:
This PR reimplements `DiskCache` to store encoded collabs in S3-compatible storage rather than in Postgres. Postgres is used only as a fallback for compatibility with an existing storage model.

TODOs:

- [x] Use S3 for collab snapshot data as well.
- [x] We need to decide if we really want to store EncodedCollab format in S3. IMO it's useless - it's heavily reliant on bincode (which is not version tolerant) and all of the fields (namely: state_vector) can be resolved from doc_state directly instead. We could keep a straight up yrs Doc state instead.
- [x] Added ZSTD compression of S3 objects.
- [x] Added a threshold (`S3_COLLAB_THRESHOLD`, 2000 bytes), so that any collab data smaller than that threshold will still land in Postgres. PS: *All new snapshots are always stored in S3 regardless of size*.


services/appflowy-worker/src/import_worker/worker.rs
```diff
@@ -1310,22 +1322,41 @@ async fn upload_file_to_s3(
...
+  let key = collab_key(workspace_id, object_id);
+  match s3.get_blob_stream(&key).await {
+    Ok(mut resp) => {
+      let mut buf = Vec::with_capacity(resp.content_length.unwrap_or(1024) as usize);
+      resp
+        .stream
+        .read_to_end(&mut buf)
+        .await
+        .map_err(|err| ImportError::Internal(err.into()))?;
+      let decompressed = zstd::decode_all(&*buf).map_err(|e| ImportError::Internal(e.into()))?;
```
appflowy, line 1340: will `get_encode_collab_from_bytes` causes tokio runtime issue If it takes 10 ms to decode the buf?

services/appflowy-worker/src/import_worker/worker.rs
```diff
@@ -1310,22 +1322,41 @@ async fn upload_file_to_s3(
...
-      "Failed to decode collab from bytes: {:?}",
-      err
-    ))),
-  })
-  .await
-  .map_err(|err| ImportError::Internal(err.into()))?
+  let key = collab_key(workspace_id, object_id);
+  match s3.get_blob_stream(&key).await {
+    Ok(mut resp) => {
+      let mut buf = Vec::with_capacity(resp.content_length.unwrap_or(1024) as usize);
```
appflowy, line 1334: would it be necessary to call `buf.try_reserve()`?

libs/database/src/collab/disk_cache.rs
```diff
@@ -121,14 +190,102 @@ impl CollabDiskCache {
...
+    let mut blobs = HashMap::new();
+    for param in params_list.iter_mut() {
+      if param.encoded_collab_v1.len() > S3_COLLAB_THRESHOLD {
+        let key = collab_key(workspace_id, &param.object_id);
+        let blob = std::mem::take(&mut param.encoded_collab_v1);
+        blobs.insert(key, blob);
+      }
+    }
+
+    let mut transaction = self.pg_pool.begin().await?;
```
appflowy, line 209: If params_list is empty, then we don't need to start a transaction

libs/database/src/collab/disk_cache.rs
```diff
@@ -73,20 +116,46 @@ impl CollabDiskCache {
...
-      query.collab_type,
-      query.object_id
-    );
+    tracing::debug!("try get {}:{} from s3", query.collab_type, query.object_id);
+    let key = collab_key(workspace_id, &query.object_id);
+    match self.s3.get_blob(&key).await {
+      Ok(resp) => {
+        let blob = resp.to_blob();
+        let now = Instant::now();
+        let decompressed = zstd::decode_all(&*blob)?;
```
appflowy, line 135: Just to make sure whether encode or decode should not stuck the tokio runtime.

Horusiath: The problem with loading the decoding to separate thread is that it involves context switching between threads, which is kernel operation and each context switch takes the kernel about 5 μs (on average) to process. And this doesn't involve time required for creating a new thread (ie. if we use rayon). That means that we add ~10 μs for every decompression task.

I'd keep it as is, and if CPU indeed will be an issue, add rayon afterwards. We can also switch to lz4 (which is faster but has worse compression sizes) to reduce these times, since the thing is not merged yet.

services/appflowy-worker/src/import_worker/worker.rs
Horusiath: If we stored the collab that's bigger than memory, it's already separate issue. But it shouldn't happen, given that we had to decode that collab through collab_group to save it in the first place.

PR #1045, opened 2024-12-05
chore: move collab cache for appflowy collaborate service and add metrics

Horusiath:
This PR:
1. Moves `CollabCache` from libs/databae to services/appflowy-collaborate.
2. Adds collab metrics tracked by cache above:
    - `(s3/pg/redis)_write_collab_count` prometheus counters to track where collabs are written.
    - `(s3/pg/redis)_read_collab_count` prometheus counters to track from where collabs are read most often
    - `pg_tx_collab_millis` histogram to measure time spent in Postgres transactions for every operation that involves writing to `af_collab` table.


services/appflowy-worker/src/import_worker/worker.rs
```diff
@@ -987,15 +985,6 @@ async fn process_unzip_file(
         err
       ))
     })?;
-    // Update the workspace database cache because newly created workspace databases are cached in Redis.
-    mem_cache
-      .insert_encode_collab(
```
Horusiath, line 992: @appflowy I think removing this part is the reason why 2 tests are failing but why? After all men_cache is just a cache. If it's missing collab lookup should fallback to postgres/s3, right?

appflowy: it's used to update existing object that was cached in redis. `insert_into_af_collab_bulk_for_user` doesn't update corresponding redis cache. `insert_into_af_collab_bulk_for_user` should not update the cache because not all objects will be used immediatly.

PR #1048, opened 2024-12-06
fix: filter non space document from workspace's children

khorshuheng:
Filter out documents which is not a space from workspace's children.

Some existing tests failed as the documents are created under a workspace as opposed to a space, which has been modified in this PR.


speed2exe (approved):
LGTM

Issue #984, opened 2024-11-12
[Bug] SMTP username isn't necessarily an email address


strongoose:
**Describe the bug**
The AppFlowy Cloud mailer uses the SMTP username to set the From address on outbound emails: https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/libs/mailer/src/sender.rs#L67

However, many (most?) SaaS email providers don't use the source email address as the SMTP username, e.g.:
 - Azure Communication services uses a string of the format `<resource name>.<app id>.<tenant id>`
 - SendGrid uses the string `apikey`

It should be possible to set the source address for AppFlowy Cloud emails separately from the SMTP credentials.

**To Reproduce**
1. Create a free tier account on SendGrid (for example)
2. Set SMTP credentials in `.env`
3. Attempt to email an invite to your AppFlowy workspace
4. AppFlowy attempts to cast the SMTP user to a `lettre` Address object, resulting in the following error:
    ```
    appflowy_cloud-1    | {"timestamp":"2024-11-12T09:50:47.600395Z","level":"ERROR","fields":{"message":"Failed to send workspace invite email: Missing domain or user\n\nStack backtrace:\n   0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from\n   1: appflowy_cloud::biz::workspace::ops::invite_workspace_members::{{closure}}::{{closure}}::{{closure}}::{{closure}}\n   2: tokio::runtime::task::raw::poll\n   3: std::sys_common::backtrace::__rust_begin_short_backtrace\n   4: core::ops::function::FnOnce::call_once{{vtable.shim}}\n   5: std::sys::pal::unix::thread::Thread::new::thread_start\n   6: <unknown>\n   7: clone"},"target":"appflowy_cloud::biz::workspace::ops","filename":"src/biz/workspace/ops.rs","line_number":461}
    ```

**Expected behavior**
It should be possible to authenticate with non-email-like SMTP credentials.

**Additional context**
 - I'm running AppFlowy Cloud using the provided docker-compose files.

speed2exe:
@strongoose 
You may now able to specify the following env var differently:
`APPFLOWY_MAILER_SMTP_USERNAME=<resource name>.<app id>.<tenant id>`
`APPFLOWY_MAILER_SMTP_EMAIL=email_sender@some_company.com`

[...]

PR #1049, opened 2024-12-06
fix: add the necessary environment variables for docker-compose.yml

owenyang0:
Related to #984 and #994.

If these environment variables are not added to the docker-compose.yml file, the workspace invitation process will fail.


khorshuheng:
Thanks!

PR #1053, opened 2024-12-07
chore: enable running stress on on self-hosted runner

appflowy:
run test that start with `stress_test` on self-hosted runner

### DIFF
--- a/deploy.env
+++ b/deploy.env
@@ -19,14 +19,9 @@ REDIS_PORT=6379
 MINIO_HOST=minio
 MINIO_PORT=9000
 
-# Gotrue Host
-GOTRUE_HOST=gotrue
-GOTRUE_PORT=9999
-GOTRUE_DATABASE_HOST=postgres
-
 # AppFlowy Cloud
 ## URL that connects to the gotrue docker container
-APPFLOWY_GOTRUE_BASE_URL=http://${GOTRUE_HOST}:${GOTRUE_PORT}
+APPFLOWY_GOTRUE_BASE_URL=http://gotrue:9999
 ## URL that connects to the postgres docker container
 APPFLOWY_DATABASE_URL=postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
 APPFLOWY_ACCESS_CONTROL=true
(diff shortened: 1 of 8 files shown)


deploy.env
```diff
@@ -15,9 +15,18 @@ SUPABASE_PASSWORD=root
...
+
+# Gotrue Host
+GOTRUE_HOST=gotrue
+GOTRUE_PORT=9999
+GOTRUE_DATABASE_HOST=postgres
+
 # AppFlowy Cloud
 ## URL that connects to the gotrue docker container
-APPFLOWY_GOTRUE_BASE_URL=http://gotrue:9999
+APPFLOWY_GOTRUE_BASE_URL=http://${GOTRUE_HOST}:${GOTRUE_PORT}
```
khorshuheng, line 29: Since AppFlowy Cloud is installed within the same docker compose network as Gotrue, are there scenarios where we need a different gotrue host?

deploy.env
```diff
@@ -66,11 +75,12 @@ GOTRUE_ADMIN_PASSWORD=password
 # If you are using a different domain, you need to change the redirect_uri in the OAuth2 configuration
 # Make sure that this domain is accessible to the user
 # Make sure no endswith /
-API_EXTERNAL_URL=http://your-host
+# Replace with your host name instead of localhost
+API_EXTERNAL_URL=http://localhost:${GOTRUE_PORT}
```
khorshuheng, line 79: This will cause the login to fail, as API_EXTERNAL_URL is used in redirect URL. In the situation where this is deployed on local machine, it should be either be localhost, or localhost:(nginx port), instead of gotrue port.

deploy.env
```diff
@@ -66,11 +75,12 @@ GOTRUE_ADMIN_PASSWORD=password
...
 
 # In docker environment, `postgres` is the hostname of the postgres service
 # GoTrue connect to postgres using this url
-GOTRUE_DATABASE_URL=postgres://supabase_auth_admin:${SUPABASE_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
+GOTRUE_DATABASE_URL=postgres://supabase_auth_admin:${SUPABASE_PASSWORD}@${GOTRUE_DATABASE_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
```
khorshuheng, line 83: Currently there is no way to run gotrue and appflowy on separate database. Hence, there is no good reason to have two separate host configuration.

appflowy: Noted

deploy.env
```diff
@@ -66,11 +70,12 @@ GOTRUE_ADMIN_PASSWORD=password
 # If you are using a different domain, you need to change the redirect_uri in the OAuth2 configuration
 # Make sure that this domain is accessible to the user
 # Make sure no endswith /
+# Replace with your host name instead of localhost
 API_EXTERNAL_URL=http://your-host
 
 # In docker environment, `postgres` is the hostname of the postgres service
 # GoTrue connect to postgres using this url
-GOTRUE_DATABASE_URL=postgres://supabase_auth_admin:${SUPABASE_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
+GOTRUE_DATABASE_URL=postgres://supabase_auth_admin:${SUPABASE_PASSWORD}@postgres:${POSTGRES_PORT}/${POSTGRES_DB}
```
khorshuheng, line 78: We do have users who prefer to use external postgres host instead of the postgres within docker. So it would be good to keep postgres_host

PR #1059, opened 2024-12-10
Remove async awaits from RealtimeServerActor

Horusiath:
This PR turns off necessity of using async message handlers in `RealtimeServerActor`. The major reason for using these were:
1. Stopping group requires calling an async `stop` method, which was using an async channel to notify group processing loop that it should be stopped. Now I've changed it to use `CancellationToken` which stops the group when the group instance is dropped. The same goes for group subscribers.
2. `CollabGroup::new` was async because of async lock over collab used inside. However we can pass a `Collab` which isn't wrapped with lock into a group, initialise everything and then wrap it with necessary lock at the end of group constructor.
3. Client message handler was async due to [this](https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/ee2ff899a3bb16a13bbabadd88a1625526c67fd2/services/appflowy-collaborate/src/rt_server.rs#L202). However the notifier here is not necessary. We're just scheduling the new messages onto group command runner while its processing loop starts on its own.


appflowy:
LGTM

PR #1064, opened 2024-12-11
chore: query embedding

appflowy:


khorshuheng:
If we perform edits only via AppFlowy Web, the collab failed to be persisted because the collab was dropped before the force save method is called, likely due to another background process that drop collab when there is no connected users.

PR #1067, opened 2024-12-12
feat: upsert database row

speed2exe:
## HTTP API to allow user to create or insert a database row based on a `pre_hash`
`pre_hash` will eventually be hashed into a `uuid` so that given the same `pre_hash`, the same row_id will be generated.
This `uuid` will then be used as `row_id` for upsert operation;

- [x] Upsert Database row
- [x] Refactor Upsert and Insert to reuse code as much as possible
- [x] Database Document Insert
- [x] Database Document Update
- [x] Tests

Non functional task
- [x] Upsert Document: check for document changes, only update doc if content is changed


src/api/workspace.rs
```diff
@@ -1960,18 +1961,69 @@ async fn post_database_row_handler(
...
+
+  let UpsertDatatabaseRow {
+    pre_hash,
+    cells,
+    document,
+  } = upsert_db_row.into_inner();
+
+  let row_id = {
+    let mut hasher = Sha256::new();
+    hasher.update(workspace_id.clone());
```
appflowy, line 2001: IIRC, the update function is `fn update(&mut self, data: impl AsRef<[u8]>)`
So you can remove clone

src/biz/collab/ops.rs
```diff
@@ -624,13 +639,158 @@ pub async fn insert_database_row(
...
+      CollabParams {
+        object_id: row_id.to_string(),
+        encoded_collab_v1: db_row_ec_v1.into(),
+        collab_type: CollabType::DatabaseRow,
+      },
+      &mut db_txn,
+      "inserting new database row from server",
+    )
+    .await?;
+  broadcast_update(collab_storage, row_id, db_row_collab_updates).await?;
```
appflowy, line 721: can we spawn broadcast_update?

src/biz/collab/ops.rs
```diff
@@ -804,5 +973,74 @@ pub async fn list_database_row_details(
...
+        object_id: doc_id.to_string(),
+        collab_type: CollabType::Document,
+      })
+      .collect::<Vec<_>>();
+
+    let mut query_res = collab_storage
+      .batch_get_collab(&uid, &workspace_uuid_str, query_db_docs, true)
+      .await;
+
+    for row_detail in &mut db_row_details {
```
appflowy, line 1003: Would it be better to create a function for the following calls? This would allow you to eliminate many match statements and take advantage of the Result type for error handling.

src/api/workspace.rs
speed2exe: done

src/biz/collab/ops.rs
speed2exe: have refactored this part into a function

PR #1068, opened 2024-12-13
feat: api for delete items from trash permanently

khorshuheng:
API for deleting items from trash permanently. Currently, this API only modifies the workspace folder, as opposed to actually delete the collab from the database. This is consistent with the behavior of the current version of AppFlowy desktop app. We will need to revisit this in the future.

This API is also differ from other page view related APIs, as the updates are written to in-memory Collab Group as opposed to directly to the database. After this PR is merged, we will migrate all the other APIs to follow similar approach.

### DIFF
--- a/src/biz/workspace/page_view.rs
+++ b/src/biz/workspace/page_view.rs
@@ -50,7 +50,8 @@ use shared_entity::dto::workspace_dto::{
 use sqlx::{PgPool, Transaction};
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
-use std::time::Instant;
+use std::time::{Duration, Instant};
+use tokio::time::timeout_at;
 use tracing::instrument;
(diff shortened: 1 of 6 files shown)


src/biz/workspace/page_view.rs
```diff
@@ -1300,3 +1370,30 @@ pub async fn update_page_collab_data(
...
+  appflowy_web_metrics.record_update_size_bytes(update.len());
+
+  let message = ClientHttpUpdateMessage {
+    user,
+    workspace_id: workspace_id.to_string(),
+    object_id: workspace_id.to_string(),
+    collab_type: CollabType::Folder,
+    update: Bytes::from(update),
+    state_vector: None,
+    return_tx: None,
```
appflowy, line 1391: If you want to wait for the applied result, then you can set Some for return_tx.

PR #1085, opened 2024-12-18
chore: find question message from reply message

richardshiue:
Currently, question (sent by user) messages in the `chat_message_table` may contain a value in the `reply_message_id` column to reference its answer message. However, answer messages do not have a reference to the question that w as asked. This PR allows for a way to try to find the question to which an answer was made.

### DIFF
--- a/libs/client-api/src/http_chat.rs
+++ b/libs/client-api/src/http_chat.rs
@@ -262,6 +262,30 @@ impl Client {
       .into_data()
   }
 
+  pub async fn get_question_message_from_answer_id(
+    &self,
+    workspace_id: &str,
+    chat_id: &str,
+    answer_message_id: i64,
+  ) -> Result<Option<ChatMessage>, AppResponseError> {
+    let mut url = format!(
+      "{}/api/chat/{workspace_id}/{chat_id}/message/find_question",
+      self.base_url
+    );
+    let query_params = vec![("answer_message_id", answer_message_id.to_string())];
+
+    let query = serde_urlencoded::to_string(&query_params).unwrap();
+    url = format!("{}?{}", url, query);
+    let resp = self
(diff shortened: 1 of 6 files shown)


libs/client-api/src/http_chat.rs
```diff
@@ -262,6 +262,30 @@ impl Client {
...
+      "{}/api/chat/{workspace_id}/{chat_id}/message/find_question",
+      self.base_url
+    );
+    let query_params = vec![("answer_message_id", answer_message_id.to_string())];
+
+    let query = serde_urlencoded::to_string(&query_params).unwrap();
+    url = format!("{}?{}", url, query);
+    let resp = self
+      .http_client_with_auth(Method::GET, &url)
+      .await?
```
appflowy, line 281: can call `.query()` before the send. In this way, you don't need to use serde_urlencoded.

src/api/chat.rs
```diff
@@ -349,6 +353,22 @@ async fn get_chat_message_handler(
   Ok(AppResponse::Ok().with_data(messages).into())
 }
 
+#[instrument(level = "debug", skip_all, err)]
+async fn get_chat_question_message_handler(
+  path: web::Path<(String, String)>,
+  query: web::Query<HashMap<String, String>>,
```
appflowy, line 359: another approach is to create a struct that derives serde::Deserialize, then you can use web::Query<YourStructName>

libs/client-api/src/http_chat.rs
richardshiue: Good to know! That's much cleaner

src/api/chat.rs
richardshiue: Thanks for this. Converting result to and from option had me scratching my head

PR #1093, opened 2024-12-19
chore: check collab type before index

appflowy:


services/appflowy-collaborate/src/indexer/indexer_scheduler.rs
```diff
@@ -209,12 +220,17 @@ impl IndexerScheduler {
...
-    indexed_collabs: Vec<IndexedCollab>,
+    mut indexed_collabs: Vec<IndexedCollab>,
   ) -> Result<(), AppError> {
     if !self.index_enabled() {
       return Ok(());
     }
 
+    indexed_collabs.retain(|collab| self.is_indexing_enabled(&collab.collab_type));
+    if indexed_collabs.is_empty() {
+      return Ok(());
```
Horusiath, line 231: The context in which this method is used means, that you can first deep copy 10 000 collabs (all of database row type => unindexable), so that you just drop them here.

services/appflowy-collaborate/src/indexer/indexer_scheduler.rs
```diff
@@ -159,8 +163,15 @@ impl IndexerScheduler {
       return Ok(());
     }
 
-    let embedder = self.create_embedder()?;
     let indexed_collab = indexed_collab.into();
+    if !self.is_indexing_enabled(&indexed_collab.collab_type) {
```
Horusiath, line 167: You can use `IndexerProvider::indexer_for` method which already return `None` if indexing is not enabled. This way you also don't need to copy indexer provider inside of rayon thread.

PR #1097, opened 2024-12-21
added a hint to warn user of GoTrue password incompability with special characters

KevinCCucumber:
Today I tried to set up appflowy cloud, but was not able to log into the admin panel.
After a few hours of debugging it dawned upon me: I had set a password containing special characters like ^!# and +.

To prevent other users from falling into the same pit, I added a warning note to the deployment .env


khorshuheng:
I think some of the special characters might be acceptable - the ones that I know would cause problem is @ , but I am not sure about ? Or #. Need to check.

khorshuheng:
Implemented in https://github.com/AppFlowy-IO/AppFlowy-Cloud/pull/1127 .

PR #1107, opened 2024-12-31
fix: auto patch workspace folder without space

khorshuheng:
In older version of AppFlowy, there's no concept of space. As a result, they can't be rendered in AppFlowy Web easily. In this PR, we will attempt to fix such case based on the following logic:

1. If the view that corresponds to the workspace has children, but none of the children are a valid space, or there is no children at all, then create a new space.
2. If there is at least one children, then all the children directly under the workspace view should be moved under the newly created space.

### DIFF
--- a/src/biz/collab/ops.rs
+++ b/src/biz/collab/ops.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
+use actix_web::web::Data;
 use app_error::AppError;
 use appflowy_collaborate::collab::storage::CollabAccessControlStorage;
 use chrono::DateTime;
@@ -24,8 +25,12 @@ use collab_database::workspace_database::WorkspaceDatabaseBody;
 use collab_document::document::Document;
 use collab_entity::CollabType;
 use collab_entity::EncodedCollab;
+use collab_folder::hierarchy_builder::NestedChildViewBuilder;
 use collab_folder::CollabOrigin;
+use collab_folder::Folder;
 use collab_folder::SectionItem;
+use collab_folder::SpaceInfo;
+use collab_rt_entity::user::RealtimeUser;
 use database::collab::select_last_updated_database_row_ids;
 use database::collab::select_workspace_database_oid;
 use database::collab::{CollabStorage, GetCollabOrigin};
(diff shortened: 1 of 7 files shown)


src/biz/collab/ops.rs
```diff
@@ -293,10 +303,86 @@ pub async fn get_user_trash_folder_views(
...
+#[allow(clippy::too_many_arguments)]
+fn patch_old_workspace_folder(
+  user: RealtimeUser,
+  workspace_id: &str,
+  folder: &mut Folder,
+  child_view_id_without_space: &[String],
+) -> Result<Vec<u8>, AppError> {
+  let encoded_update = {
+    let space_id = Uuid::new_v4().to_string();
+    let view = NestedChildViewBuilder::new(user.uid, workspace_id.to_string())
```
appflowy, line 315:
```rust
 pub fn space_info(&self) -> Option<SpaceInfo> {
    let extra = self.extra.as_ref()?;
    serde_json::from_str::<SpaceInfo>(extra).ok()
  }
```
in collab repo, we have a method call space_info(). So we need to insert this space info in the view extra

LucasXu0: I think you can use `space_info` instead of building the extra value manually.

src/biz/collab/ops.rs
```diff
@@ -293,10 +303,86 @@ pub async fn get_user_trash_folder_views(
...
+      })
+      .build()
+      .view;
+    let mut txn = folder.collab.transact_mut();
+    folder.body.views.insert(&mut txn, view, None);
+    for view_id in child_view_id_without_space {
+      folder
+        .body
+        .move_nested_view(&mut txn, view_id, &space_id, None);
+    }
```
LucasXu0, line 332:
```rs
    // associate the child with its new parent and place it after the prev_view_id. If the prev_view_id is None,
    // place it as the first child.
    self
      .views
      .associate_parent_child_with_txn(txn, new_parent_id, view_id, prev_view_id.clone());
```

I'm not sure if the order is handled correctly. The order should be preserved after the move.

khorshuheng:
Depends on https://github.com/AppFlowy-IO/AppFlowy-Collab/pull/362 . Once the PR is merged i will update AppFlowy-Collab version to use the commit on the main branch instead of the PR branch.

PR #1108, opened 2024-12-31
feat: api for publish page to web

khorshuheng:
API endpoint for publishing a page to web.


src/biz/workspace/page_view.rs
```diff
@@ -1034,6 +1041,216 @@ pub async fn update_page(
   Ok(())
 }
 
+fn replace_invalid_url_chars(input: &str) -> String {
+  let regex = Regex::new(r"[^\w-]").unwrap();
```
appflowy, line 1045: we can use lazy_static here

PR #1109, opened 2024-12-31
chore: add prometheus histogram metrics for collab redis update read latency

Horusiath:


services/appflowy-collaborate/src/metrics.rs
```diff
@@ -112,8 +122,22 @@ impl CollabRealtimeMetrics {
...
+      "collab_stream_latency",
+      "latency since collab update is send to a stream to be read from it",
+      metrics.collab_stream_latency.clone(),
+    );
     metrics
   }
+
+  pub fn observe_collab_stream_latency(&self, message_id_timestamp: u64) {
+    let now = Utc::now().timestamp_millis() as u64;
+    if now > message_id_timestamp {
```
khorshuheng, line 135: Under what circumstances will the message timestamp to be greater than the current time?

Horusiath: If clocks between Redis and appflowy-cloud processes are desynchronided, NTP clock correction happens for any reason or if the leaps seconds happens.

khorshuheng:
Lgtm

PR #1111, opened 2025-01-01
fix: added missing env parameter for ADMIN_FRONTEND_APPFLOWY_CLOUD_URL to deploy.env

nicolaus-hee:
`docker-compose.yml` has env variable `ADMIN_FRONTEND_APPFLOWY_CLOUD_URL` set to `${ADMIN_FRONTEND_APPFLOWY_CLOUD_URL:-http://appflowy_cloud:8000}`, though `deploy.env` makes no mention of the variable. In my environment, I needed to change the hostname which would also require changing it in `deploy.env` where it is not mentioned. I suggest adding it with the default value for convenience and consistency (the other two `ADMIN_FRONTEND_...` are included) purposes.

### DIFF
--- a/deploy.env
+++ b/deploy.env
@@ -35,6 +35,8 @@ APPFLOWY_REDIS_URI=redis://${REDIS_HOST}:${REDIS_PORT}
 ADMIN_FRONTEND_REDIS_URL=redis://${REDIS_HOST}:${REDIS_PORT}
 ## URL that connects to gotrue docker container
 ADMIN_FRONTEND_GOTRUE_URL=http://gotrue:9999
+## URL that connects to the cloud docker container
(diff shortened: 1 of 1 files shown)


khorshuheng:
We can (and certainly should) add this variable. Although, if gotrue service and appflowy cloud service are deployed on the same network/docker compose, gotrue service can reach appflowy cloud directly via the service name `appflowy_cloud`, unless you change the service name for appflowy cloud.

khorshuheng:
Anyway, this PR looks good to me. Once the commit message has been amended, we can merge this. The commit message should be in the form of (fix: <description starting with lower case>)

khorshuheng (approved):

PR #1116, opened 2025-01-03
chore: add metrics to redis collab stream

Horusiath:
This PR comes with 2 changes:
1. `CollabStreamMetrics` type which allows us to track how many streams are we scheduling to be read at given time:
  - `collab_stream_reads_enqueued` is incremented for every stream that has been observed. When the Redis xread returns and stream needs to be rescheduled for the next poll iteration, it also becomes enqueued.
  - `collab_stream_reads_dequeued` is incremented for every stream that scheduler returned for xread. These two metrics should be fairly close to each other. If they reads_enqueued grows faster than reads_dequeued it means, that we schedule new stream read requests faster than we can handle them.
2. It fixes the issue when an observer was not correctly dropped and stayed inside of stream router for as long as no new objects where pulled from its corresponding stream.

### DIFF
--- a/libs/collab-stream/src/stream_router.rs
+++ b/libs/collab-stream/src/stream_router.rs
@@ -1,3 +1,4 @@
+use crate::metrics::CollabStreamMetrics;
 use loole::{Receiver, Sender};
 use redis::streams::{StreamReadOptions, StreamReadReply};
 use redis::Client;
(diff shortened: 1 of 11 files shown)


libs/collab-stream/src/stream_router.rs
```diff
@@ -198,21 +213,27 @@ impl Worker {
...
     senders: &mut HashMap<&str, (StreamSender, usize)>,
-  ) {
+  ) -> usize {
     let keys = keys.drain(..);
     let mut ids = ids.drain(..);
+    let mut scheduled = 0;
     for key in keys {
       if let Some(last_id) = ids.next() {
         if let Some((sender, _)) = senders.remove(key.as_str()) {
+          if sender.is_closed() {
```
Horusiath, line 223: The problem was when there were no new messages in Redis stream. In such case we didn't check if any active subscriber was actually requesting for them. If subscriber was closed we still tried to read the stream even if nobody wanted to listen to it.

This could mean that if we had 5000 subscribers in the past, but no new data was coming from their Redis streams, and only 5 of them where active atm. we still tried to XREAD 5000 streams instead of 5.

khorshuheng (approved):

PR #1119, opened 2025-01-03
Document paragraphs

Horusiath:
TODOs:

- [x] 1. Split documents by paragraphs, not arbitrary length (see: https://github.com/AppFlowy-IO/AppFlowy-Collab/pull/361)
- [x] 2. Create `af_collab_embeddings.fragment_id` using a consistent hash of chunk it relates to.
- [ ] 3. At the beginning of indexing, get collab content with list of its currently indexed fragment_ids.
- [ ] 4. When generating embeddings skip embeddings for those chunks which fragment_id was found in the point above and keep their content unchanged.
- [ ] 5. When persisting embeddings don't delete all of the existing ones. Delete only those that were not found in point 3, those for which we didn't got embeddings stay the same, and insert all of the new ones.


appflowy:
> 3. At the beginning of indexing, get collab content with list of its currently indexed fragment_ids.
>  4. When generating embeddings skip embeddings for those chunks which fragment_id was found in the point above and keep their content unchanged.
>  5. When persisting embeddings don't delete all of the existing ones. Delete only those that were not found in point 3, those for which we didn't got embeddings stay the same, and insert all of the new ones.

These three options will be implemented in another PR

PR #1128, opened 2025-01-05
fix: api external url should be for gotrue, not appflowy cloud

khorshuheng:
API_EXTERNAL_URL is meant for the external endpoint for gotrue, not appflowy cloud. Not having gotrue suffix will cause SAML integration to fail.


speed2exe:
If we want to remove `gotrue` prefix, we need also to modify nginx routing so that every supported path in gotrue service is routed properly. This will also break the dev env (`AppFlowyCloud Develop`) in appflowy flutter app. 
In addition, documentations in `doc/AUTHENTICATION.md` will also need to be updated.

There might also be other issues, these are the concerns I can think of at the moment.

speed2exe (approved):

PR #1131, opened 2025-01-06
feat: endpoint for unpublish page

khorshuheng:
The current endpoint, which the desktop application used to unpublish page (/workspace/{workspace_id}/publish), cannot be used by AppFlowy Web because this endpoint wrongly expects a request body for a delete method. This PR add a different endpoint that adheres to the REST API convention.

The PR also fixed a previously undiscovered bug, where unpublished pages are still considered published.


speed2exe (approved):
LGTM

PR #1141, opened 2025-01-08
fix: select most recent non original namespace

khorshuheng:
If there are multiple non original namespace, we should choose the most recent one instead of the first.


speed2exe (approved):
LGTM

PR #1144, opened 2025-01-08
fix: nginx conf to allow localhost:3000 to access api endpoint

khorshuheng:
Fix the current nginx conf to properly allow localhost:3000 to access /api endpoint.


speed2exe (approved):
LGTM

PR #1159, opened 2025-01-13
feat: add root view id of the created web template to duplicate published page response

khorshuheng:
Previously, there is no need for returning the newly created page id, because there's no requirement to jump to the duplicated page once the API is called.

However, AppFlowy Web will soon have such requirement, hence the root view id of the duplicate view is required in the response now.


speed2exe (approved):
LGTM

PR #1163, opened 2025-01-14
fix: trim awareness when collab group is dropped

khorshuheng:
Trim awareness when collab group is dropped, as there is no need to persist awareness when no users are connected. If we don't trim the awareness, when the server is restarted, all the awareness info will be read from the stream again. In addition, the size of the awareness stream will also be much more than necessary (reaching the maximum, which is 100).

This does introduce some complication for having multiple replica of AppFlowy Cloud, however, if the same collab can have collab group in two different replica. When that happen, we may have to store some information on Redis to determine whether it is safe to trim the stream. Or, alternatively, we can periodically set TTL to the Redis stream key when a Collab Group is active. If none of the groups are active in any of the replica, the key will automatically be dropped.

Also removed the minimum timeout as a criteria for removing CollabGroup - there's no reason to keep collab group when the number of subscriber reach zero.


khorshuheng:
@Horusiath FYI

appflowy (approved):

PR #1165, opened 2025-01-15
chore: include default services into extras

thuvh:
it easier to use docker-compose-extras


khorshuheng:
Thanks, could you update the commit message to follow the following format:
"chore: include default services into extras" ?

In addition, can you update this section (https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/doc/DEPLOYMENT.md#4-optional-services) , so that users will know that docker compose extra will also include docker compose?

khorshuheng (approved):

PR #1170, opened 2025-01-17
fix: do not merge awareness sync data info client update sync data

LucasXu0:
```rs
  fn receive_local_state(
    &self,
    origin: &CollabOrigin,
    object_id: &str,
    _event: &Event,
    update: &AwarenessUpdate,
  ) {
    let payload = Message::Awareness(update.encode_v1()).encode_v1();
    // ...
      ClientCollabMessage::new_update_sync(update_sync) <- Here
```

In the previous implementation, the awareness data was associated with the `ClientCollabMessage::new_update_sync`, which caused the awareness update to be merged into the client update.

### DIFF
--- a/libs/client-api/src/collab_sync/collab_sink.rs
+++ b/libs/client-api/src/collab_sync/collab_sink.rs
@@ -274,6 +274,15 @@ where
   }
 
   async fn process_next_msg(&self) {
+    let is_empty_queue = self
+      .message_queue
+      .try_lock()
+      .map(|q| q.is_empty())
+      .unwrap_or(true);
+    if is_empty_queue {
+      return;
+    }
+
     let items = {
       let (mut msg_queue, mut sending_messages) = match (
         self.message_queue.try_lock(),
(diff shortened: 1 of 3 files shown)


libs/client-api/src/collab_sync/collab_sink.rs
```diff
@@ -357,19 +366,34 @@ where
...
+        // 2. The last message can be merged and the next message can be merged.
         // 3. The last message's payload size is less than the maximum payload size.
         if let Some(last) = items.last_mut() {
-          if !sending_messages.contains(&last.msg_id())
+          let can_merge = !sending_messages.contains(&last.msg_id())
             && last.message().payload_size() < self.config.maximum_payload_size
             && last.mergeable()
-            && last.merge(&next, &self.config.maximum_payload_size).is_ok()
-          {
+            && next.mergeable()
```
LucasXu0, line 375: When merging the last and next, **we should check if the next is mergeable**. If the last is mergeable but the next is NOT, they should not be merged.

libs/client-api/src/collab_sync/plugin.rs
```diff
@@ -192,7 +192,7 @@ where
         trace!("queue awareness: {:?}", update);
       }
 
-      ClientCollabMessage::new_update_sync(update_sync)
+      ClientCollabMessage::new_awareness_sync(update_sync)
```
LucasXu0, line 195: @Horusiath could you please confirm whether it was a typo or intentional to set the `awareness update` as `update_sync` instead of `awareness_sync`?

khorshuheng:
lgtm

PR #1199, opened 2025-01-26
fix: typo in env.sh

SilvioMessi:


khorshuheng:
Thanks for catching this mistake. Somehow, the image still works despite this. Will have to find out why.

khorshuheng:
Ok, i think it works because AF_GOTRUE_URL was still set correctly. But functionality involving AF_BASE_URL might fail.

PR #1268, opened 2025-03-10
chore: replace the all `docker-compose.yml` with `compose.yaml`

Marukome0743:
## Description

There are some [`docker-compose.yml`](https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/docker-compose.yml) files.

However, according to [official manuals](https://docs.docker.com/compose/intro/compose-application-model/#the-compose-file),

> The default path for a Compose file is `compose.yaml` (preferred) or `compose.yml` that is placed in the working directory. Compose also supports `docker-compose.yaml` and `docker-compose.yml` for backwards compatibility of earlier versions. If both files exist, Compose prefers the canonical `compose.yaml`.

Therefore, it's good to use `compose.yaml` instead of `docker-compose.yml`.


almereyda:
Can we go for `compose.yml`, which is shorter and sits more lightly reachable on the keyboard?

Moving the left most finger from y to a on regular layouts is a more complicated move than just tapping y m and l.

It would also show some continuity with the original intent of amending the `.yml` suffix to `docker-compose`.

PR #1270, opened 2025-03-10
fix: update maizzle README.md

Marukome0743:
## Description

The [email_template README.md](https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/email_template/README.md) has broken links.

previous

I obtained the official README.md from maizzle and applied.

new

### DIFF
--- a/email_template/README.md
+++ b/email_template/README.md
@@ -10,29 +10,39 @@
   <p>Quickly build HTML emails with Tailwind CSS</p>
 <div>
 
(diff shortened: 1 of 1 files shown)


khorshuheng:
Maizzle is actually not owned by us - it is perhaps misleading for us to include it in the readme.

The steps in Development is actually AppFlowy specific (we define some custom script in package.json), so it should not be overwritten by the Maizzle official guide.

PR #1282, opened 2025-03-18
[WIP] Sync protocol v2

Horusiath:


khorshuheng:
We have https://github.com/AppFlowy-IO/AppFlowy-Cloud/tree/main/libs/collab-rt-entity . Could we reuse that, or consolidate that library into the new module which you created in this PR?

Cargo.toml
```diff
@@ -27,12 +27,12 @@ serde_repr.workspace = true
 serde.workspace = true
 
 tokio = { workspace = true, features = [
-  "macros",
-  "rt-multi-thread",
-  "sync",
-  "fs",
-  "time",
-  "full",
+    "macros",
```
appflowy, line 30: We recommend using two spaces as format. Otherwise, these files will be changed very frenquently

libs/client-api/src/v2/actor.rs
```diff
@@ -0,0 +1,802 @@
...
+    last_message_id: &Rid,
+    cancel: CancellationToken,
+  ) -> anyhow::Result<Option<WsConn>> {
+    let url = format!("{}/{}", options.url, options.workspace_id);
+    tracing::info!("establishing WebScoket connection to: {}", url);
+    let mut req = url.into_client_request()?;
+    let headers = req.headers_mut();
+    headers.insert("X-AF-Device-ID", HeaderValue::from_str(&options.device_id)?);
+    headers.insert(
+      "X-AF-Client-ID",
```
appflowy, line 738: IIRC, we need to update the nginx config to allow custom headers

libs/database-entity/src/dto.rs
```diff
@@ -105,13 +106,21 @@ impl PendingCollabWrite {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Validate)]
 pub struct CollabParams {
-  #[serde(with = "uuid_str")]
   pub object_id: Uuid,
-  #[validate(custom(function = "validate_not_empty_payload"))]
-  pub encoded_collab_v1: Bytes,
   pub collab_type: CollabType,
+  #[validate(length(min = 2))]
```
appflowy, line 121: Is len 2 the minimum SV size?

libs/database/src/collab/collab_db_ops.rs
```diff
@@ -194,27 +199,56 @@ pub async fn insert_into_af_collab_bulk_for_user(
...
+  let records = sqlx::query_as!(
+    AFCollabData,
+    r#"
+        SELECT c.oid, c.partition_key, c.updated_at, c.blob
+        FROM af_collab c
+        WHERE c.workspace_id = $1
+            AND c.deleted_at IS NULL
+            AND c.created_at > $2
+        ORDER BY updated_at
+        LIMIT 500
```
appflowy, line 220: Why are we selecting 500? Shouldn’t we use the value passed in by the caller? What happens if those 500 rows are actual database records?

services/appflowy-collaborate/src/group/group_init.rs
```diff
@@ -1034,7 +1030,7 @@ impl CollabPersister {
       collab.transact_mut().apply_update(update).map_err(|err| {
```
appflowy, line 1030: Would it be better to acquire a transaction once and reuse it, rather than obtaining a new one for each update?

services/appflowy-collaborate/src/ws2/actors/session.rs
```diff
@@ -0,0 +1,294 @@
...
+          },
+          ws::Message::Continuation(item) => {
+            match item {
+              Item::FirstText(_) => ctx.close(Some(CloseReason::from((
+                CloseCode::Unsupported,
+                "only binary content is supported",
+              )))),
+              Item::FirstBinary(bytes) => self.buf = Some(bytes.into()),
+              Item::Continue(bytes) => {
+                if let Some(ref mut buf) = self.buf {
```
appflowy, line 203: Same question applies to unlimited message sizes: what happens if the buffer size is too large? It could cause an OOM error.

services/appflowy-collaborate/src/ws2/actors/workspace.rs
```diff
@@ -0,0 +1,446 @@
...
+  }
+}
+
+impl Handler<Join> for Workspace {
+  type Result = ();
+
+  fn handle(&mut self, msg: Join, ctx: &mut Self::Context) -> Self::Result {
+    if msg.workspace_id == self.workspace_id {
+      let handle = SessionHandle::new(msg.collab_origin, msg.addr.clone());
+      self.sessions_by_client_id.insert(msg.session_id, handle);
```
appflowy, line 336: How should we handle this case? When a client connects twice, the first connection overrides the second—so the server continues using the first connection, while the client treats the second connection as the active one.

services/appflowy-collaborate/src/ws2/actors/workspace.rs
```diff
@@ -0,0 +1,446 @@
...
+              flags: msg.update_flags,
+              last_message_id: msg.last_message_id,
+              update: msg.update.clone(),
+            },
+          });
+        }
+      },
+      Err(err) => {
+        tracing::error!(
+          "failed to read update stream message for workpsace {}: {:?}",
```
appflowy, line 297: typo: should be workspace

services/appflowy-collaborate/src/ws2/actors/workspace.rs
```diff
@@ -0,0 +1,446 @@
...
+    }
+  }
+}
+
+impl Handler<Leave> for Workspace {
+  type Result = ();
+
+  fn handle(&mut self, msg: Leave, ctx: &mut Self::Context) -> Self::Result {
+    if msg.workspace_id == self.workspace_id {
+      self.sessions_by_client_id.remove(&msg.session_id);
```
appflowy, line 371: Should we remove the client after a timeout? The server might not receive a Leave message even if the client exits.

libs/client-api/src/v2/actor.rs
```diff
@@ -0,0 +1,802 @@
...
+            }
+          }
+        }
+        _ = keep_alive.tick() => {
+          // we didn't receive any message for some time, so we send a ping to keep connection alive
+          let actor = match weak_ref.upgrade() {
+            Some(actor) => actor,
+            None => break, // controller dropped
+          };
+          if let Err(err) = actor.ping().await {
```
appflowy, line 213: If I understand correctly, each time a ping fails we set the status to “Disconnected.” If the client immediately attempts to reconnect upon seeing that status, it’s very easy to trigger repeated reconnects, which will cause the server to receive numerous Join messages.

services/appflowy-collaborate/src/ws2/actors/workspace.rs
```diff
@@ -0,0 +1,446 @@
...
+              message: ServerMessage::Manifest {
+                object_id: msg.object_id,
+                collab_type,
+                last_message_id: Rid::default(),
+                state_vector: StateVector::default().encode_v1(),
+              },
+            });
+          },
+          Err(err) => {
+            tracing::error!(
```
appflowy, line 121: Do we need to inform the client with specific error?

services/appflowy-collaborate/src/ws2/actors/session.rs
appflowy: NVM. I saw the `frame_size(10 * 1024 * 1024)` limit

services/appflowy-collaborate/src/ws2/actors/workspace.rs
Horusiath: `insert` replaces old value (connection) with a new one and returns older if it existed - here we just drop it - so the previous connection handle is dropped. Also on the client side, there can be only one active connection at the time.

services/appflowy-collaborate/src/ws2/actors/workspace.rs
Horusiath: Why? Leave is sent by heartbeat interval or when the session actor is stopping.

libs/client-api/src/v2/actor.rs
Horusiath: Failure to send a ping, means that the connection is already broken, so it cannot be used to send any more messages reliably anyway.

The exponentiall backoff reconnect strategy should be responsibility of the one calling the reconnects, which would mean the client.

services/appflowy-collaborate/src/ws2/actors/workspace.rs
Horusiath: But what can client do about it?
1. We cannot close the connection, since it's shared between multiple collabs and this error is collab specific.
2. Error can be caused by temporary issues on redis/postgres/s3 and potentially be fixed in the future - but we can already do that when we open a collab or detect missing updates.

There's not much client can do about the server error here.

libs/database/src/collab/collab_db_ops.rs
Horusiath: I can change it into a variable, but the problem stays - we don't want to overdo a number of collabs to be returned, since we could easily flood the user.

services/appflowy-collaborate/src/ws2/actors/workspace.rs
appflowy: Network connectivity on mobile devices can be unstable. We monitor the network state, and when it transitions from disconnected to connected, we initiate a WebSocket (ws) connection. However, this transition can occur multiple times in quick succession, causing the client to attempt multiple ws connections. I believe the server should prioritize the most recent connection. If I remember correctly, the server actor doesn't guarantee the order in which connection requests (joins) are processed.

libs/client-api/src/v2/README.md
```diff
@@ -0,0 +1,929 @@
...
+- State vector tracking to identify precisely what updates are missing
+- Two-way synchronization on reconnection
+- The `WorkspaceController` client implementation that manages connection state
+
+## Resilience Mechanisms
+
+The AppFlowy collaboration system implements several mechanisms to ensure reliability and consistency across various
+challenging scenarios.
+
+### Resource ID (Rid) Management
```
Horusiath, line 308: Actually Rid stands for Redis ID. It's the same as `MessageId` we already have, but my problem was that MessageId is defined in module that is not shared dependency - also MessageId is very generic name ;)

libs/client-api/src/v2/README.md
```diff
@@ -0,0 +1,929 @@
...
+    - `handle_client_collab_message` validates the message and user
+    - `forward_message_to_group` routes it to the appropriate CollabGroup
+    - `handle_protocol_message` processes the message based on its type
+    - The update is applied to the server's document
+5. The server broadcasts the update to all other connected clients of that document
+6. Each receiving client:
+    - Applies the update to their local document
+    - Updates their state vector and last_message_id
+
+### Offline Editing & Reconnection
```
Horusiath, line 232: Tbh. there's no need to describe 3 flows with diagrams. In practice it's a single flow of messages, regardless of where the document is present or what was the connection state before. Because of that this protocol description is bigger than it actually needs.

libs/client-api/src/v2/README.md
```diff
@@ -0,0 +1,929 @@
...
+            Ordering::Equal => self.seq_no.cmp(&other.seq_no),
+            ordering => ordering,
+        }
+    }
+}
+```
+
+#### Detailed Rid Implementation and Purpose
+
+The Rid mechanism is a critical component for maintaining causal consistency in the distributed CRDT system. It provides
```
Horusiath, line 358: No, it's not. It's useful since all of the updates from all the clients come through the same Redis stream, so they are totally ordered. This is why we don't need to track causal consistency (via i.e. state vectors) when trying to read updates from the Redis Stream.

libs/client-api/src/v2/README.md
```diff
@@ -0,0 +1,929 @@
...
+     fn cmp(&self, other: &Self) -> Ordering {
+       match self.timestamp.cmp(&other.timestamp) {
+         Ordering::Equal => self.seq_no.cmp(&other.seq_no),
+         ordering => ordering,
+       }
+     }
+   }
+   ```
+
+3. **Causal History Tracking**: By tracking the last received Rid (`last_message_id`), the system can:
```
Horusiath, line 384: Rid is useful for 2 reasons:
1. So that we can quickly check for new updates if client has disconnected for a second.
2. Since it has a timestamp component, we can also check in the database if new collabs were created since the last sync time (last Rid received).

libs/client-api/src/v2/README.md
```diff
@@ -0,0 +1,929 @@
...
+       old.clone()
+     }
+   });
+   ```
+
+#### Rid in Server-Side Implementation
+
+In the server implementation (`services/appflowy-collaborate`), Rids play a crucial role:
+
+1. **Update Ordering**: The `GroupCommandRunner` in the server uses Rids to process updates in the correct order:
```
Horusiath, line 444: There's no `GroupCommandRunner` in sync protocol V2.

libs/client-api/src/v2/README.md
```diff
@@ -0,0 +1,929 @@
...
+    Note over Server: Remove ClientA from active participants
+```
+
+In this scenario:
+
+1. Client connects to the server and establishes a WebSocket connection
+2. Client sends a Manifest message for a document the server doesn't know about
+3. Server processes the client manifest:
+   - Checks for document existence and finds it doesn't exist
+   - Creates a new CollabGroup for this document
```
Horusiath, line 187: There's no `CollabGroup` in sync protocol v2.

libs/client-api/src/v2/README.md
```diff
@@ -0,0 +1,929 @@
...
+   ```
+
+#### Benefits of Rid-based Ordering
+
+- **Global Ordering**: Provides total ordering of updates across all clients
+- **Idempotency**: Duplicate update detection by tracking last_message_id
+- **Out-of-order Handling**: Later updates are recognized even if they arrive first
+- **Reconnection Efficiency**: Clients request updates newer than their last_message_id
+- **Causality Preservation**: Updates are applied in a causally consistent order
+- **Conflict Resolution Support**: Rid ordering helps resolve conflicts when merged
```
Horusiath, line 489: The AI really went overboard here ;) 1st point is saying about total ordering, while 5ft is going on about causal order. Pretty much none of these points has any sense.

libs/client-api/src/v2/README.md
```diff
@@ -0,0 +1,929 @@
...
+
+The Rid mechanism, combined with CRDT state vectors, forms the foundation of AppFlowy's robust synchronization system,
+providing consistent document states even in challenging network conditions.
+
+### Disconnection and Reconnection
+
+When a client disconnects, the system:
+
+- Keeps local changes in memory and continues tracking the state vector
```
Horusiath, line 498: Not only in memory but also on disk.

libs/client-api/src/v2/README.md
```diff
@@ -0,0 +1,929 @@
...
+```rust
+async fn handle_disconnect(&self) -> anyhow::Result<()> {
+    let previous_status = self
+        .status_tx
+        .send_replace(ConnectionStatus::Disconnected { reason: None });
+    // ... disconnect logic
+}
+```
+
+The reconnection process is managed by a dedicated `ReconnectionManager` that:
```
Horusiath, line 511: There's no such thing as `ReconnectionManager` as far as I know (at least not in this repo).

Horusiath (commented):
I read through half of the generated documentation markdown and my key point is that we probably should not use AI to write the documentation :D

libs/client-api/src/v2/README.md
appflowy: I implemented it

libs/client-api/src/v2/actor.rs
```diff
@@ -305,16 +324,26 @@ impl WorkspaceControllerActor {
...
     } = &msg
     {
       let rid = source.into();
+      let changed_collab = ChangedCollab {
+        id: *object_id,
+        collab_type: *collab_type,
+      };
+      // Only insert if not already present
+      if !self.latest_changed_collabs.contains(&changed_collab) {
+        self.latest_changed_collabs.insert(changed_collab);
```
Horusiath, line 365: Why do you need to check if it's there before insertion?

libs/client-api/src/v2/db.rs
```diff
@@ -129,24 +131,31 @@ impl Db {
...
       return Ok(None);
     }
     let instance = self.inner.get()?;
     let ops = instance.write_txn();
+
+    #[cfg(feature = "verbose_log")]
     tracing::trace!(
-      "persisting update for {}/{} by {}",
+      "persisting {} update for {}/{} by {}. update: {:#?}",
+      action_source,
```
Horusiath, line 169: If the only reason for `action_source` to be passed as a parameter is to be logged inside of the method, maybe it's better to put the log line before this method is called?

libs/client-api/src/v2/actor.rs
```diff
@@ -76,6 +81,20 @@ impl WorkspaceControllerActor {
     actor
   }
 
+  /// Return current changed collab ids and clear the set.
+  pub async fn consume_latest_changed_collabs(&self) -> HashSet<ChangedCollab> {
```
Horusiath, line 85: 1. This method doesn't have to be asynchronous.
2. Why use map? If you want to track the changed collabs, you could use `tokio::sync::broadcast::channel` and just consume it with await using `recv` or `try_recv`.

appflowy: flutter client periodically batch consume changed collabs. Any benefit by using channel for this case?

libs/client-api/src/v2/actor.rs
appflowy:
```
if !self.latest_changed_collabs.contains(object_id) {
        self.latest_changed_collabs.insert(ChangedCollab {
          id: *object_id,
          collab_type: *collab_type,
        });
      }
```
Contains is cheaper than insert

PR #1283, opened 2025-03-19
chore: awareness pubsub switch to lightweight version

Horusiath:
This PR changes `AwarenessGossip` implementation to reduce a number of connections created. I suspect that current main implementation creates a 2 redis connections per collab. This one uses only 2 connections total, regardless of number of collabs.

### DIFF
--- a/libs/collab-stream/src/awareness_gossip.rs
+++ b/libs/collab-stream/src/awareness_gossip.rs
@@ -1,28 +1,53 @@
 use crate::error::StreamError;
 use crate::model::AwarenessStreamUpdate;
-use async_stream::try_stream;
-use futures::Stream;
+use dashmap::DashMap;
 use redis::aio::MultiplexedConnection;
-use redis::{AsyncCommands, Client};
+use redis::{AsyncCommands, Client, RedisError};
+use std::sync::Arc;
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 use tokio::sync::Mutex;
 use tokio_stream::StreamExt;
+use uuid::Uuid;
 
 pub struct AwarenessGossip {
-  client: Client,
+  conn: MultiplexedConnection,
+  collabs: Arc<DashMap<Uuid, UnboundedSender<AwarenessStreamUpdate>>>,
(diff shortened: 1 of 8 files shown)


libs/collab-stream/src/awareness_gossip.rs
```diff
@@ -1,75 +1,83 @@
...
-
-  pub fn state_key(workspace_id: &str, object_id: &str) -> String {
-    format!("af:awareness_state:{}:{}", workspace_id, object_id)
-  }
+  pub async fn new(client: &Client) -> Result<Self, RedisError> {
+    let collabs: Arc<DashMap<Uuid, UnboundedSender<AwarenessStreamUpdate>>> =
+      Arc::new(DashMap::new());
+    let mut pub_sub = client.get_async_pubsub().await?;
+    pub_sub.psubscribe("af:awareness:*").await?;
+    let conn = client.get_multiplexed_async_connection().await?;
```
Horusiath, line 23: We only create 2 connections here:
1. One used for pattern subscribe on all `af:awareness:*` channels - these will be pulled by separate task and dispatched in memory to corresponding collabs.
2. One multiplexed connection for sending publish messages on a given channel. Multiplexed connection uses Redis pipelinening and can be cloned without reconnecting to make a new connection.

libs/collab-stream/src/awareness_gossip.rs
```diff
@@ -1,75 +1,83 @@
...
 
-  pub fn awareness_stream(
-    &self,
-    workspace_id: &str,
-    object_id: &str,
-  ) -> impl Stream<Item = Result<AwarenessStreamUpdate, StreamError>> + 'static {
-    let client = self.client.clone();
-    let pubsub_key = Self::publish_key(workspace_id, object_id);
-    try_stream! {
-        let mut pubsub = client.get_async_pubsub().await?;
```
Horusiath, line 51: This could also create a new connection as well.

appflowy:
LGTM

appflowy (approved):

khorshuheng (approved):

PR #1286, opened 2025-03-20
Migrate af_collab to non-partitioned table with UUID keys

Horusiath:


migrations/20250318120849_departition_af_collab.sql
```diff
@@ -0,0 +1,103 @@
...
+    created_at timestamp with time zone default CURRENT_TIMESTAMP,
+    updated_at timestamp with time zone default CURRENT_TIMESTAMP not null,
+    indexed_at timestamp with time zone
+);
+
+-- move non-UUID object ids to the new table (execution time: 31 secs)
+insert into af_collab_conflicts(oid, workspace_id, owner_uid, partition_key, len, blob, deleted_at, created_at, updated_at, indexed_at)
+select oid, workspace_id, owner_uid, partition_key, len, blob, deleted_at, created_at, updated_at, indexed_at
+from af_collab
+where oid !~ E'^[[:xdigit:]]{8}-([[:xdigit:]]{4}-){3}[[:xdigit:]]{12}$';
```
Horusiath, line 37: Since we don't want to remove user data, we move all incompatible records to separate table. Operators will have to resolve these issues.

khorshuheng (approved):

PR #1298, opened 2025-03-25
chore: init env filter by parsing the default env variable

khorshuheng:
Log level for different crate can be controlled directly using RUST_LOG, such as RUST_LOG=appflowy_cloud=info,appflowy_collaborate=trace. This should be controlled by env variable rather than hard coded within application.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,19 +6,6 @@ use appflowy_cloud::telemetry::init_subscriber;
 async fn main() -> anyhow::Result<()> {
   let level = std::env::var("RUST_LOG").unwrap_or("info".to_string());
   println!("AppFlowy Cloud with RUST_LOG={}", level);
-  let mut filters = vec![];
-  filters.push(format!("actix_web={}", level));
-  filters.push(format!("collab={}", level));
-  filters.push(format!("collab_sync={}", level));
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -6,27 +6,14 @@ use appflowy_cloud::telemetry::init_subscriber;
...
-  filters.push(format!("collab_sync={}", level));
-  filters.push(format!("appflowy_cloud={}", level));
-  filters.push(format!("collab_plugins={}", level));
-  filters.push(format!("realtime={}", level));
-  filters.push(format!("database={}", level));
-  filters.push(format!("storage={}", level));
-  filters.push(format!("gotrue={}", level));
-  filters.push(format!("appflowy_collaborate={}", level));
-  filters.push(format!("appflowy_ai_client={}", level));
-  filters.push(format!("indexer={}", level));
```
Horusiath, line 21: We need to be wary how these changes will affect test/beta deployments.

khorshuheng: In production environment, we have defined RUST_LOG=info, and the same applies to self hosters too, for the default deployment. We can adjust RUST_LOG as necessary.

Horusiath (approved):

PR #1331, opened 2025-04-11
fix: update ai environment variables

Marukome0743:
## Summary

At #1133, some ai environment variables are renamed.
This PR updates them and add the missing ones in [docker-compose.yml](https://github.com/AppFlowy-IO/AppFlowy-Cloud/blob/main/docker-compose.yml).


docker-compose.yml
```diff
@@ -174,9 +174,18 @@ services:
     image: appflowyinc/appflowy_ai:${APPFLOWY_AI_VERSION:-latest}
     environment:
       - OPENAI_API_KEY=${AI_OPENAI_API_KEY}
-      - APPFLOWY_AI_SERVER_PORT=${AI_SERVER_PORT}
-      - APPFLOWY_AI_DATABASE_URL=${AI_DATABASE_URL}
-      - APPFLOWY_AI_REDIS_URL=${AI_REDIS_URL}
+      - ANTHROPIC_API_KEY=${AI_ANTHROPIC_API_KEY}
```
khorshuheng, line 177: This would actually cause the current AI container to fail, because the service itself still expects the old variable name, instead of the updated one.

khorshuheng:
While the variables were renamed in the PR, they are applicable to the private AppFlowy AI image used by our managed service, rather than the public AppFlowy AI image which is available for self hosters. You would see, that the image on docker compose CI is actually different from the one in docker compose. One has the premium suffix, the other one doesn't.

Nevertheless, these inconsistencies are indeed confusing. So we might update the current public AppFlowy AI image to adopt the same env variable as the premium variant.

PR #1390, opened 2025-05-14
chore: implement custom prompt usage metrics

richardshiue:
## Summary by Sourcery

Add support for tracking and recording custom prompt usage metrics by introducing a prompt_usage_count metric, extending DTOs with an optional prompt_id, updating API handlers to log usage, and updating tests to cover the new field.

New Features:
- Track custom prompt usage count per prompt ID in metrics

Enhancements:
- Add optional prompt_id field to chat and AI client DTOs
- Update API handlers to record prompt usage metrics when prompt_id is provided

Tests:
- Update AI client completion tests to include prompt_id in test data


khorshuheng:
Given that we don't have a large number of prompt id, and not all of them will be used by the users, Prometheus might still be able to handle the cardinality. 

Prometheus isn't really made for analytical metrics, though, so we will need to reconsider the current approach if the number of prompt ids grow. cc: @appflowy

khorshuheng (approved):

PR #1423, opened 2025-06-02
chore: use rayon to write snapshot updates

appflowy:
## Summary by Sourcery

Parallelize collab snapshot generation and saving using Rayon, introduce a dedicated thread pool, distributed locking, and batch processing to improve throughput and avoid conflicts.

Enhancements:
- Add a ThreadPoolNoAbort in CollabStore for parallel snapshot processing
- Convert sequential snapshot loop into a Rayon-powered parallel pipeline
- Implement a Redis-based distributed lock to prevent concurrent workspace snapshots
- Batch snapshot saving and indexing in chunks to manage resource usage
- Extract process_collab_with_snapshot helper to encapsulate collab state reconstruction

Build:
- Add futures and nanoid crates to dependencies

Tests:
- Add placeholder tests/main.rs

### DIFF
--- a/services/appflowy-collaborate/src/collab/cache/collab_cache.rs
+++ b/services/appflowy-collaborate/src/collab/cache/collab_cache.rs
@@ -1,5 +1,7 @@
 use super::disk_cache::CollabDiskCache;
 use super::mem_cache::{cache_exp_secs_from_collab_type, CollabMemCache};
+use crate::collab::cache::DECODE_SPAWN_THRESHOLD;
+use crate::config::get_env_var;
 use crate::CollabMetrics;
 use app_error::AppError;
 use appflowy_proto::{Rid, UpdateFlags};
@@ -27,9 +29,6 @@ use yrs::updates::decoder::Decode;
 use yrs::updates::encoder::Encode;
 use yrs::{ReadTxn, StateVector, Update};
 
-/// Threshold for spawning background tasks for memory cache operations.
-const MEM_CACHE_SPAWN_THRESHOLD: usize = 4096; // 4KB
-
 #[derive(Clone)]
 pub struct CollabCache {
   thread_pool: Arc<ThreadPoolNoAbort>,
@@ -39,7 +38,7 @@ pub struct CollabCache {
   /// Threshold for spawning background tasks for memory cache operations.
   /// Data smaller than this will be processed on the current thread.
   /// Data larger than this will be spawned to avoid blocking.
-  mem_cache_spawn_threshold: usize,
+  small_collab_size: usize,
   metrics: Arc<CollabMetrics>,
 }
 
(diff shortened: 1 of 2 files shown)


libs/client-api/src/v2/actor.rs
```diff
@@ -521,7 +531,9 @@ impl WorkspaceControllerActor {
     while let Some(res) = stream.next().await {
       if cancel.is_cancelled() {
         sync_trace!("remote receiver loop cancelled");
-        break;
+        return Err(DisconnectedReason::MessageLoopEnd(
+          "remote receiver loop cancelled".into(),
```
Horusiath, line 535: Same here: cancel is called by the client on explicit disconnection. It's not an error.

services/appflowy-collaborate/src/collab/cache/collab_cache.rs
```diff
@@ -196,48 +210,74 @@ impl CollabCache {
...
+    // Determine the state vector to use for generating the diff:
+    // 1. If provided in the query, use it directly
+    // 2. Otherwise, try to decode from the existing encoded collab
+    // 3. Fall back to empty state vector if decoding fails or no collab exists
+    let from = from.unwrap_or_else(|| {
+      encoded_collab
+        .as_ref()
+        .and_then(|ec| match ec.version {
+          EncoderVersion::V1 => StateVector::decode_v1(&ec.state_vector).ok(),
+          EncoderVersion::V2 => StateVector::decode_v2(&ec.state_vector).ok(),
```
Horusiath, line 227: Do we ever even try to encode state vectors using v2 encoding?

services/appflowy-collaborate/src/collab/cache/collab_cache.rs
```diff
@@ -196,48 +210,74 @@ impl CollabCache {
...
-      {
-        let mut tx = collab.transact_mut();
-        for msg in updates {
-          if msg.object_id == object_id {
-            let update = match msg.update_flags {
-              UpdateFlags::Lib0v1 => Update::decode_v1(&msg.update),
-              UpdateFlags::Lib0v2 => Update::decode_v2(&msg.update),
+      encoded_collab = self
+        .thread_pool
+        .install(|| {
```
Horusiath, line 235: What I'm worried about is that for small collabs (ie. DatabaseRows), process of calling install may take longer than the decoding+encoding. Maybe we could add some metrics to thing call?

services/appflowy-collaborate/src/ws2/collab_store.rs
```diff
@@ -390,7 +430,61 @@ pub struct CollabState {
   pub state_vector: Vec<u8>,
 }
 
-fn decode_update(update: &[u8]) -> AppResult<Update> {
+pub fn process_collab_with_snapshot(
```
Horusiath, line 433: Better name? `process_` doesn't describe anything. Maybe `apply_updates_to_snapshot`?

services/appflowy-collaborate/src/collab/cache/collab_cache.rs
appflowy: Yes. I will fallback to previous approach.

PR #1434, opened 2025-06-06
chore: check if collab was recently edited before fetching updates from Redis stream

Horusiath:
## Summary by Sourcery

Implement a caching optimization that tracks recently edited collabs via a dirty flag to avoid unnecessary Redis stream reads, marking collabs as dirty on update and cleaning them after flushing to disk.

Enhancements:
- Introduce a dirty_collabs set in CollabCache to track objects with pending updates
- Add mark_as_dirty and mark_as_clean methods to manage dirty collabs
- Skip Redis stream fetch for collabs that aren’t marked dirty, returning cached state directly
- Mark collabs as dirty when publishing or persisting updates across CollabStore, GroupInit, CollabPersister, and workspace actor
- Extend the CollabStorage trait with a mark_as_editing method

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -355,7 +355,7 @@ pub async fn init_state(config: &Config, rt_cmd_tx: CLCommandSender) -> Result<A
   let collab_store = CollabStore::new(
     thread_pool.clone(),
     collab_storage_access_control.clone(),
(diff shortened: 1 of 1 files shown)


appflowy:
@Horusiath LGTM, but some tests were failed

PR #1496, opened 2025-06-30
chore: remove workspace redis stream on workspace deletion

Horusiath:
## Summary by Sourcery

Clean up Redis streams for workspaces upon deletion by passing the Redis connection manager through the user and workspace deletion flows and issuing a DEL command for the workspace stream key.

Enhancements:
- Forward the Redis connection manager through delete_user and delete_workspace_for_user functions
- Delete the Redis stream key for a workspace in delete_workspace_for_user

### DIFF
--- a/src/biz/workspace/ops.rs
+++ b/src/biz/workspace/ops.rs
@@ -12,8 +12,9 @@ use tracing::instrument;
 use uuid::Uuid;
 
 use access_control::workspace::WorkspaceAccessControl;
-use app_error::AppError;
+use app_error::{AppError, ErrorCode};
 use appflowy_collaborate::CollabMetrics;
+use collab_stream::model::UpdateStreamMessage;
 use database::collab::CollabStore;
 use database::file::s3_client_impl::S3BucketStorage;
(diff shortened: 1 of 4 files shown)


src/biz/workspace/ops.rs
```diff
@@ -53,6 +55,10 @@ pub async fn delete_workspace_for_user(
 
   // remove from postgres
   delete_from_workspace(&pg_pool, &workspace_id).await?;
+  let _: redis::Value = connection_manager
+    .del(UpdateStreamMessage::stream_key(&workspace_id))
+    .await
+    .map_err(|err| AppResponseError::new(ErrorCode::Internal, err.to_string()))?;
```
appflowy, line 61: how is the client handle this error? Maybe we just need to log it on the server side.

Horusiath: Tbh. I think we could just retry here. I don't expect it to fail for any other reason than network issues.

appflowy: I've checked the client side implementation.  it should be fine

PR #1521, opened 2025-07-17
feat: s3 multipart import file upload

strophy:
This PR introduces support for large file (multipart) imports in AppFlowy-Cloud in order to work around the [5GB limitation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/upload-objects.html) in AWS S3 on single PUT operations.

It unifies the logic for small and large file uploads, updating both API and internal handling. The changes ensure that files larger than 5GB are uploaded using a multipart protocol compatible with S3 and the AppFlowy worker, while maintaining backward compatibility for small files.

I have successfully used this to import a 12.2GB Notion export in AWS EC2 with S3 storage and an external proxy. I have a corresponding PR [modifying AppFlowy-Web here](https://github.com/AppFlowy-IO/AppFlowy-Web/pull/134). Some further changes to proxy template defaults may be needed? I also haven't tested or modified the desktop AppFlowy client because I don't use it, but I guess changes are needed here too to support this function. There shouldn't be any breaking changes though, so the old desktop client should still work.

I'm looking forward to edits and CI results, hopefully my approach wasn't too naive here, my understanding of AppFlowy is still very limited.

## Summary by Sourcery

Implement S3-compatible multipart import file upload for files over 5GB and unify the import upload workflow between small and large files

New Features:
- Support multipart upload for large import files in client-api with upload_large_import_file
- Extend create_import endpoint to choose between presigned URL and multipart upload and return upload_type and workspace_id in the response

Enhancements:
- Unify upload_import_file method to delegate to small or large upload logic based on file size

Build:
- Add tempfile as a dev-dependency for creating temporary test files

Tests:
- Add tests for small single-part and large multipart import file uploads
- Update import test helper to supply workspace_id for multipart uploads


khorshuheng:
If i understand correctly, for this approach, the server will need to have sufficient disk space / memory in order to handle the file upload, since the file will be uploaded indirectly to S3 via the server instead of using presigned url directly.

This is fine (and a good way to get around the large file limitation imposed by S3) for self hosted use cases, as the server will typically have sufficient disk / memory for a single person.

But, when there are large number of users, the server will require quite a bit of resource, and may crash if multiple users are trying to upload the files at the same time.

PR #1637, opened 2026-08-01
fix(deploy): add runtime admin authentication fallback

tonicofonico:
### Description

Add a Docker-runtime option for using browser-side authentication instead of the Admin frontend's authentication Server Actions:

```env
APPFLOWY_ADMIN_DISABLE_SERVER_ACTIONS=false
```

The option is documented in `deploy.env` and passed to `admin_frontend` by Docker Compose. It remains disabled by default.

### Problem

Behind some TLS-terminating reverse proxies, Server Action requests can lose the expected cookie or origin context and cause an authentication redirect loop.

The original version of this PR used `NEXT_PUBLIC_DISABLE_SERVER_ACTIONS`. That name is a build-time Next.js setting, so setting it when starting the prebuilt Admin image does not update the browser bundle.

### Runtime behavior

This PR is paired with [AppFlowy-Admin #67](https://github.com/AppFlowy-IO/AppFlowy-Admin/pull/67), which injects `APPFLOWY_ADMIN_DISABLE_SERVER_ACTIONS` into `window.__APP_CONFIG__`:

- unset or `false`: continue using authentication Server Actions
- `true`: use the existing browser-side authentication API fallback

This setting only changes Admin authentication behavior; it does not globally disable Next.js Server Actions.

### Dependency

AppFlowy-Admin #67 must be merged and a compatible `appflowyinc/admin_frontend` image must be published before this PR is ready to merge.

### Verification

- `docker compose --env-file deploy.env config --quiet`
- verified Compose renders the default as `false` and an explicit override as `true`
- built the Admin image from #67 and started it through this Compose service definition
- verified the container receives `APPFLOWY_ADMIN_DISABLE_SERVER_ACTIONS=true`
- verified `/console/login` exposes the runtime value to browser code


appflowy:
Maintainer follow-up: I pushed `581c77d5` to replace the ineffective runtime `NEXT_PUBLIC_DISABLE_SERVER_ACTIONS` setting with `APPFLOWY_ADMIN_DISABLE_SERVER_ACTIONS`. The original `NEXT_PUBLIC_*` value is frozen when the Admin image is built and was not visible to browser code when set by Compose.

